PR #20, opened 2015-03-22
Stan egs import

bob-carpenter:
This is a bunch of models, all tested, brought over from various Stan branches.


syclik:
thanks, Bob.

PR #22, opened 2015-06-02
Small corrections to Rate Comparison Example from Lee & Wagenmakers 

singmann:
This commit contains mainly a tiny addition to the rate comparison example of Lee & Wagenmakers (2014) using the Geurts data (Chapter 9.4). The code was missing the prior for effect size `delta`. This prior was present in all other examples, but not in this example in file `Greuts_Stan.R`. 

In addition to this error correction I added some cosmetic changes to the code and renamed the files from the incorrect spelling `Greuts` to the correct spelling `Geurts`.


bob-carpenter:
@signmann thanks!
@syclik  do you know what's up with Travis CI on this?

syclik:
I'm not sure we set anything up. Should we?

neekro:
@singmann Thanks, much appreciated!

PR #32, opened 2015-08-20
ARM ch3 model updates

ericnovik:


bob-carpenter:
The changes to the Stan models all look OK to me.  I didn't try to follow the R wrappers.

It does bring up the point as to how closely we want to follow the models in the book.  Do we want to have the direct translation and then what we currently think is the right way to do it?

I think you need Andrew's feedback on this --- maybe you can grab him before or during the next meeting --- he's prone to have meetings scheduled starting an hour after the Stan meeting starts, so don't dally.

Maybe you could help Andrew get started on the Wiki we created for him that he's been asking for to list all the priors he wants to use.  It's a moving target, so I'm not sure what to do about the gazillions of models we have out there when Andrew changes his mind about which priors he wants to use.  Just please don't do it while other people are talking (or take Andrew out of the room) because it's impossible to hear remotely when there are side conversations during our meetings.

ericnovik:
Thanks for reviewing. I will check with Andrew during the next meeting.
The R code changes are minor and mostly consist of reducing the number of
iterations, including pairs() plots, and fixing relative path names.

bob-carpenter:
This is OK to me, so we might as well merge to make it better than it used to be rather than waiting.

PR #37, opened 2015-12-18
BPA models

ito4303:
These are Stan translated models from "Bayesian Population Analysis using WinBUGS". In Twitter, @mcmc_stan invited me to publish the models. Now I have completed the translation up to Chapter 8.


bob-carpenter:
Thanks so much for submitting them.  We went over them briefly in the Stan meeting today and will send you some more comments inline.   The biggest problems with the models we see (assuming they are translated properly from the book) are:

* sampling statements not vectorized
* missing upper bounds on scale parameters (usually <code>sigma</code>) that have uniform distributions

For the second point,

```
parameters {
  real<lower=0> sigma;
  ...
model {
   sigma ~ uniform(0, 5);
   ...
```

is not consistent --- Stan requires support for any variable value that matches the constraints.  Here, <code>sigma=10</code> matches the constraints, but doesn't have support in the model.  Although we strongly advise against uniform priors on scale parameters, if that's what you want to match the book, you need to declare

```
parameters {
  real<lower=0, upper=5> sigma;
  ..
```

instead.

If you can fix all the models for these two issues, we can review the code in more depth.

ito4303:
Thank you for your comments. I will fix the models for these issues.

ito4303:
I have fixed the models.
- sampling statements are vectorized if possible
- scale parameters now have upper boundaries

I keep priors on scale parameters matching the book.

bob-carpenter:
Sorry to be so slow reviewing this.  Our last reviewer stalled out and gave up because of the scope of reviewing required.  

One issue for me is that the merge is too big to actually display in GitHub.  Rayleigh pointed me to the instructions you get from "command line instructions" under merge:

```
git checkout -b ito4303-BPA master
git pull https://github.com/ito4303/example-models.git BPA
```

So i managed to check it out and will have it reviewed by next week.  Thanks again for submitting all this!

bob-carpenter:
Usually we don't commit files that are produced automatically by the code in the package.   It's a great sanity check for code reviewing.

@ito4303 : Do you really think it's necessary to have all the R output?  It's not a huge deal one way or the other.

Also, the R outputs are hard to read because they wrap --- it'd probably be better to include

```
options("width"=120)
```

in the R scripts or whatever's necessary so they don't wrap.

ito4303:
Thank you for reviewing my submission.

I agree that the R output files are not necessary.  So, I removed Rout files from the commit.

BPA/Ch.03/GLM_Poisson.stan
```diff
@@ -0,0 +1,38 @@
...
+
+  log_lambda <- alpha +
+                beta1 * year +
+                beta2 * year .* year +
+                beta3 * year .* year .* year;
+}
+
+model {
+  // Priors
+  alpha ~ uniform(-20, 20);
```
bob-carpenter: When parameters get uniform priors, they should be declared with matching constraints.  The rule is that every parameter value matching the declared constraints should have a finite log density (not -infinity).  So you need `<lower=-20,upper=20>` on `alpha`.

BPA/Ch.04/GLM0.stan
```diff
@@ -0,0 +1,39 @@
...
+parameters {
+  real alpha;
+}
+
+transformed parameters {
+  real log_lambda[nyear, nsite];
+
+  for (i in 1:nyear)
+    for (j in 1:nsite)
+      log_lambda[i, j] <- alpha;
```
bob-carpenter: I don't see why `log_lambda` is being defined as `alpha` here.  Why not just use `alpha` in the model directly?

BPA/Ch.04/GLM1.stan
```diff
@@ -0,0 +1,38 @@
...
+    for (j in 1:nsite)
+      log_lambda[i, j] <- alpha[j];
+}
+
+model {
+  // Priors
+  alpha ~ normal(0, 10);    // Site effects
+
+  for (i in 1:nobs)
+    obs[i] ~ poisson_log(log_lambda[obsyear[i], obssite[i]]);
```
bob-carpenter: Why not replace the `log_lambda[...]` with `alpha[obssite[i]]`?

BPA/Ch.04/GLMM1.stan
```diff
@@ -0,0 +1,41 @@
...
+}
+
+model {
+  // Priors
+  alpha ~ normal(mu_alpha, sd_alpha);
+  mu_alpha ~ normal(0, 10);
+  sd_alpha ~ uniform(0, 5);
+
+  for (i in 1:nobs)
+    obs[i] ~ poisson_log(log_lambda[obsyear[i]][obssite[i]]);
```
bob-carpenter: Using a comma here would match the earlier code.

BPA/Ch.04/GLM0.stan
ito4303: This is a direct translation from the original code of the book. I think that the reason is for expandability to incorporate year effects and/or site effects.

BPA/Ch.04/GLM1.stan
ito4303: This is same as above. The authors seem to prefer consistency rather than efficiency.

bob-carpenter: OK, makes sense.

BPA/Ch.03/GLM_Binomial.stan
```diff
@@ -0,0 +1,39 @@
...
+
+transformed parameters {
+  vector[nyears] logit_p;
+
+  // Linear predictor
+  logit_p <- alpha +
+             beta1 * year +
+             beta2 * year .* year;
+}
+
```
bob-carpenter: I think I lost my last comment like this, but here's another case.

Given that `year` is data, you want to compute `year .* year` as transformed data and assigned it to a new vector `year_squared`.  And then use `beta2 * year_squared* instead of `beta2 * year .* year` --- it saves a lot of compution to move operations to transformed data when possible.  Here, Stan is left associative, so that the long form parses as `(beta2 * year) .* year`, which introduces a lot of unnecessary autodiff expressions compared to the single multiplication by `year_squared`.

BPA/Ch.06/Mth.stan
```diff
@@ -0,0 +1,76 @@
...
+      // z[i] == 1
+      increment_log_prob(bernoulli_log(1, omega) +
+                         bernoulli_log(y[i], p[i]));
+    } else { // s[i] == 0
+      // z[i] == 1
+      lp[1] <- bernoulli_log(1, omega) +
+               bernoulli_log(0, p[i]);
+      // z[i] == 0
+      lp[2] <- bernoulli_log(0, omega);
+      increment_log_prob(log_sum_exp(lp));
```
bob-carpenter: There's a two-argument log_sum_exp that's a tiny bit more efficient, but not worth changing here.

bob-carpenter:
This all looks awesome.  Thanks so much.  It's all ready to merge.

Just one last request.  Would you mind adding a LICENSE.txt file with either the contents of or a link to the new-BSD or GPLv3 license in it?  You'll maintain copyright and ownership of the code, but we need a license before we can distribute your work.

Also, I'd like to announce this on Andrew Gelman's blog --- do you have a web link you'd like me to use other than your GitHub page?

ito4303:
Thank you for reviewing. I fixed the issues that you point out.

I also added LICENSE.txt with the contents of new-BSD license.
My web link is <http://researchmap.jp/read0208767/>.

Thank you again for your cooperation.

jgabry:
These look great. Thanks for doing all this work!

bob-carpenter:
I merged it, linked to the models from our top-level
README, and scheduled a blog post on Andrew's blog for
tomorrow (Thursday, 1 PM NY Time) to announce them.

Thanks again.

- Bob

PR #38, opened 2015-12-22
numerical typos in ARM examples

tjmahr:
I'm working through the Gelman/Hill book using the ARM examples. I've caught a couple of typos (numbers that don't match the text of the book).


rayleigh:
Thanks for pointing out these inconsistencies and for correcting them. We've merged them in.

PR #39, opened 2016-01-18
Issue 34 case study egs

bob-carpenter:
Add simple tutorial for hierarchical models for repeated binary trials.  Knitr formatted with data and simulation included.


bob-carpenter:
OK, took everyone's comments into account, so I'm going to merge.

PR #41, opened 2016-02-03
knitr pooling R code edits

jgabry:
@bob-carpenter I was going over this again in order to see if I could fit these models (or similar ones) in rstanarm and I couldn’t help myself from making a few changes to the R code. I figured I’d do this as a pull request so you can either decide you want to use these changes and merge it or just close the pull request if you prefer to leave the code as is.

Summary: 

* re-use a single function for all of the similar scatterplots

* use greek letters in axis labels where appropriate

* simplify data creation and plotting code for bda_5_4 figure (the plot of observed vs estimated success rate with 80% intervals)

* remove one section where the same data frame is created twice

### DIFF
--- a/knitr/pool-repeated-trials/pool-no-pool.Rmd
+++ b/knitr/pool-repeated-trials/pool-no-pool.Rmd
@@ -1014,50 +1014,32 @@ intervals for the estimated chance-of-success parameters $\theta_n$ in
 the posterior.  The following R code reproduces a similar plot for our data.
 
 ```{r}
-ss_quantile <- function(ss, N, q) {
-  result <- rep(NA, N);
-  for (n in 1:N) {
-    result[n] <- sort(ss$theta[,n])[M * q];
-  }
-  return(result);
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks.  I'm always of two minds writing tutorials on how much to allow myself to cut-and-paste.  On the one hand, it makes the examples less verbose and points out their commonalities, whic is good.  On the other hand, it makes them more indirect and less self-contained, which is bad.

I'm just going to merge this because you probably have a better sense of how to teach people R than I do!

jgabry:
On Wed, Feb 3, 2016 at 1:19 PM, Bob Carpenter <notifications@github.com>
wrote:

> Thanks. I'm always of two minds writing tutorials on how much to allow
> myself to cut-and-paste. On the one hand, it makes the examples less
> verbose and points out their commonalities, whic is good. On the other
> hand, it makes them more indirect and less self-contained, which is bad.
>
Yeah I agree. It's hard to find the right balance.

PR #43, opened 2016-02-12
BPA: Add Chapter 9

ito4303:
I added models translated from Chapter 9 of BPA.


BPA/Ch.09/ladyslipper.stan
```diff
@@ -0,0 +1,126 @@
...
+      ps[3, i, t, 2] <- s[t] * psiD[2];
+      ps[3, i, t, 3] <- s[t] * psiD[3];
+      ps[3, i, t, 4] <- 1.0 - s[t];
+      ps[4, i, t, 1] <- 0.0;
+      ps[4, i, t, 2] <- 0.0;
+      ps[4, i, t, 3] <- 0.0;
+      ps[4, i, t, 4] <- 1.0;
+
+      // Define probabilities of O(t) given S(t)
+      po[1, i, t, 1] <- 1.0;
```
bob-carpenter: po should be defined as transformed data.  And if you could manage it somehow, so should ps[4].  

And you can use the new syntax to do something like  

```
po[,i,t,] <- diag_matrix(rep_vector(1, 3));
```

but it'd be even better to just define a single unit matrix and reuse it (that is, drop the middle indices, which don't do anything).

BPA/Ch.09/ladyslipper.stan
```diff
@@ -0,0 +1,126 @@
...
+  b ~ gamma(1, 1);
+  c ~ gamma(1, 1);
+
+  // Likelihood
+  // Forward algorithm derived from Stan Modeling Language
+  // User's Guide  and Reference Manual
+  for (i in 1:nind) {
+    if (first[i] > 0) {
+      for (k in 1:4)
+        gamma[first[i], k] <- if_else(y[i, first[i]] == k, 1.0, 0.0);
```
bob-carpenter: This can just be this (whenever you see `if_else(c, 1, 0)` or `if_else(c, 0, 1)` it can be made logical).

```
gamma[first[i], k] <- (y[i, first[i]] == k);
```

BPA/Ch.09/ms3_multinomlogit.stan
```diff
@@ -0,0 +1,145 @@
...
+  for (i in 1:2) {
+    psiA[i] <- exp(lpsiA[i]) / (1.0 + exp(lpsiA[1]) + exp(lpsiA[2]));
+    psiB[i] <- exp(lpsiB[i]) / (1.0 + exp(lpsiB[1]) + exp(lpsiB[2]));
+    psiC[i] <- exp(lpsiC[i]) / (1.0 + exp(lpsiC[1]) + exp(lpsiC[2]));
+  }
+  // Calculate the last transition probability
+  psiA[3] <- 1.0 - psiA[1] - psiA[2];
+  psiB[3] <- 1.0 - psiB[1] - psiB[2];
+  psiC[3] <- 1.0 - psiC[1] - psiC[2];
+
```
bob-carpenter: You can use `softmax` to define these with a function defined as

```
vector softmax_1(vector lp) {
  vector[num_elements(lp) + 1] lp_temp;
  lp_temp[1:num_elements(lp)] <- lp;
  lp_temp[num_elements(lp) + 1] <- 0;
  return softmax(lp_temp);
}
```

BPA/Ch.09/ms_alternative1.stan
```diff
@@ -0,0 +1,115 @@
...
+    po[3, t, 3] <- 1.0;
+  } //t
+}
+
+model {
+  real acc[3];
+  vector[3] gamma[n_occasions];
+
+  // Priors
+  mean_phi ~ uniform(0, 1);
```
bob-carpenter: You don't need the uniforms when the parameter has the appropriate constraints.  And you need those constraints.  But if you think it makes it clearer having it, feel free to leave it in --- it's not that expensive to do the bounds check each iteration, which is the only extra overhead.

BPA/Ch.09/ms_alternative1.stan
```diff
@@ -0,0 +1,115 @@
...
+  mean_psi ~ uniform(0, 1);
+  mean_p ~ uniform(0, 1);
+
+  // Likelihood
+  // Forward algorithm derived from Stan Modeling Language
+  // User's Guide and Reference Manual
+  for (i in 1:nind) {
```
bob-carpenter: If this is the same in all the programs, it might be clearer to pull it out as a function so its shared structure is apparent.

bob-carpenter:
Looks great --- this is going to be super useful to ecologists.

I only had a few comments that are more about efficiency and numerical stability than the models themselves.  Feel free to fix what you want and then just drop a note when you'd like me to merge it.

BPA/Ch.09/ms_alternative1.stan
ito4303: I leave this statement in concordance with the original BUGS code.

BPA/Ch.09/ms_alternative1.stan
ito4303: I could not do so because an increment_log_prob function is inside the loop and it must be in the model block.

BPA/Ch.09/ms3_multinomlogit.stan
ito4303: Thank you for your suggestion. I incorporated the function into the model. In addition, I added comments on functions as the section C.8 of Stan reference manual recommends.

BPA/Ch.09/ms_alternative1.stan
bob-carpenter: You can use functions that end in `_lp` and then call them in the model block and they can access `increment_log_prob()` internally.  It's really a matter of preference given that we don't support libraries of functions yet.

ito4303:
Thank you for reviewing and merging the code.

bob-carpenter:
It's so well organized and consistent that it's a pleasure!

I took a look at the upcoming chapters of Kery's book.  There's
an example of Dorazio and Royle's occupancy model in 
stan-dev/example-models/knitr.

PR #44, opened 2016-02-19
Add education case studies and 2PL tutorial

danielcfurr:
Adding subfolder for education research examples.


bob-carpenter:
This is perfect.  Thanks.

PR #46, opened 2016-02-26
BPA: Add Chapter 10

ito4303:
I translated models in Chapter 10 of BPA. Would you please review the code?


BPA/Ch.10/js_ms.stan
```diff
@@ -0,0 +1,156 @@
...
+      cprob[t] <- gamma[t] * prod(qgamma[1:(t - 1)]);
+    } //t
+    psi <- sum(cprob);
+    for (t in 1:(n_occasions - 1)) {
+      b[t] <- cprob[t] / psi;
+    } //t
+
+    for (i in 1:M) {
+      for (t in 2:n_occasions) {
+        if (z[i, t] == 2)
```
bob-carpenter: This can just be

al[i, t - 1] <- (z[i, t] == 2);

BPA/Ch.10/js_ms.stan
```diff
@@ -0,0 +1,156 @@
...
+      } //t
+      alive[i] <- sum(al[i]);
+    } //i
+
+    for (t in 1:(n_occasions - 1)) {
+      N[t] <- sum(al[1:M, t]);
+      B[t] <- sum(d[1:M, t]);
+    } //t
+    for (i in 1:M) {
+      if (alive[i] == 0)
```
bob-carpenter: And this

```
w[i] <- (alive[i] != 0);
```

BPA/Ch.10/js_rest_occ.stan
```diff
@@ -0,0 +1,232 @@
...
+
+    for (i in 1:nind) {
+      chi[i, n_occasions] <- 1.0;
+      for (t in 1:(n_occasions - 1)) {
+        int t_curr;
+        int t_next;
+
+        t_curr <- n_occasions - t;
+        t_next <- t_curr + 1;
+        chi[i, t_curr] <- (1 - phi[i, t_curr]) +
```
bob-carpenter: I have a very strong preference for operators beginning lines so that code's easy to scan down the left.  Ideally, each additive term gets its own line.  So that'd be:

```
chi[i, t_curr] <- (1 - phi[i, t_curr])
    + phi[i, t_curr] * (1 - p[i, t_next]) *  chi[i, t_next];
```

BPA/Ch.10/js_rest_occ.stan
```diff
@@ -0,0 +1,232 @@
...
+
+    if (first[i]) { // Captured
+      // Until first capture
+      if (first[i] == 1) {
+        1 ~ bernoulli(gamma[1] * p[i, 1]);
+      } else {
+        vector[first[i]] lp;
+
+        // Entered at 1st occasion
+        lp[1] <- bernoulli_log(1, gamma[1]) +
```
bob-carpenter: I think it's clearer with addition symbols starting the line (you can't do that in R, but you can in Stan).

BPA/Ch.10/js_rest_occ.stan
```diff
@@ -0,0 +1,232 @@
...
+        recruit[i, t] <- (1 - z[i, t - 1]) * z[i, t];
+      }
+    }
+    for (t in 1:n_occasions) {
+      N[t] <- sum(z[1:M, t]);
+      B[t] <- sum(recruit[1:M, t]);
+    }
+    for (i in 1:M) {
+      Nind[i] <- sum(z[i]);
+      Nalive[i] <- int_step(Nind[i]);
```
bob-carpenter: This can just be:

```
Nalive[i] <- (Nind[i] > 0);
```

I find that clearer than `int_step`.

BPA/Ch.10/js_super.stan
```diff
@@ -0,0 +1,247 @@
...
+
+    for (i in 1:nind) {
+      chi[i, n_occasions] <- 1.0;
+      for (t in 1:(n_occasions - 1)) {
+        int t_curr;
+        int t_next;
+
+        t_curr <- n_occasions - t;
+        t_next <- t_curr + 1;
+        chi[i, t_curr] <- (1.0 - phi[i, t_curr]) +
```
bob-carpenter: Same comment as above

BPA/Ch.10/js_super.stan
```diff
@@ -0,0 +1,247 @@
...
+      1 ~ bernoulli(psi);
+
+      // Until first capture
+      if (first[i] == 1) {
+        1 ~ bernoulli(nu[1] * p[i, 1]);
+      } else {  // first[i] >= 2
+        vector[first[i]] lp;
+
+        // Entered at 1st occasion
+        lp[1] <- bernoulli_log(1, nu[1]) +
```
bob-carpenter: Same again --- plus signs should begin lines.

BPA/Ch.10/js_super_indran.stan
```diff
@@ -0,0 +1,255 @@
...
+
+model {
+  vector[n_occasions] qnu;
+
+  qnu <- 1.0 - nu;
+
+  // Priors
+  mean_phi ~ uniform(0, 1);
+  mean_p ~ uniform(0, 1);
+  for (i in 1:M)
```
bob-carpenter: This should be vectorized to

```
epsilon ~ normal(0, sigma);
```

And we don't like those interval constraints as you've declared for `epsilon` --- it'd be better to just leave it unconstrained; the hierarchical prior should control the scale.

BPA/Ch.10/js_tempran.stan
```diff
@@ -0,0 +1,239 @@
...
+model {
+  vector[n_occasions] qgamma;
+
+  qgamma <- 1.0 - gamma;
+
+  // Priors
+  mean_phi ~ uniform(0, 1);
+  mean_p ~ uniform(0, 1);
+  gamma ~ uniform(0, 1);
+  for (t in 1:(n_occasions - 1))
```
bob-carpenter: Vectorize as

```
epsilon[1 : (n_occassions - 1)] ~ normal(0, sigma);
```

BPA/Ch.10/js_tempran.stan
```diff
@@ -0,0 +1,239 @@
...
+
+    if (first[i]) { // Captured
+      // Until first capture
+      if (first[i] == 1) {
+        1 ~ bernoulli(gamma[1] * p[i, 1]);
+      } else {
+        vector[first[i]] lp;
+
+        // Entered at 1st occasion
+        lp[1] <- bernoulli_log(1, gamma[1]) +
```
bob-carpenter: Pluses begin lines.

bob-carpenter:
Thanks!  These look great.  I have some very minor stylistic comments inline and one comment on interval-based priors, which we're trying to remove from our examples.  If you want to do exactly what the book models say, could you put a comment in the code that we wouldn't recommend the interval constraints?

ito4303:
Thank you for reviewing. I fixed comment styles and removed the interval constraint.

BPA/Ch.10/js_super_indran.stan
```diff
@@ -0,0 +1,254 @@
...
+    last[i] <- last_capture(y[i]);
+}
+
+parameters {
+  real<lower=0,upper=1> mean_phi;             // Mean survival
+  real<lower=0,upper=1> mean_p;               // Mean capture
+  real<lower=0,upper=1> psi;                  // Inclusion probability
+  vector<lower=0>[n_occasions] beta;
+  vector[M] epsilon;
+  real<lower=0,upper=5> sigma;
```
bob-carpenter: Here's another one where we'd rather just have a prior.  The problem with intervals is that if the data is consistent with values near the boundaries, it is hard to compute and the posterior mean and variance will be affected.

bob-carpenter:
I spotted one more in the final pass.  When that's cleaned up, I'll go ahead and merge.  And thanks again for doing such a great job coding all these.

BPA/Ch.10/js_super_indran.stan
ito4303: The declaration is changed to `real<lower=0,upper=20> sigma;`. The posterior 97.5% quantile of sigma^2 was 3.259.

bob-carpenter: Is there some reason you don't want to just get rid of the upper=20
constraint?  If you want to leave it in for fidelity to the book,
we can do that (though I'd put a comment in to the effect that we
don't recommend it in Stan and it's only there because the authors
of the book did it).

The issue isn't so much where the 0.975 quantile is at, but whether
upper=5 (or upper=20) leads to different posterior means and variances.
Imagine a Cauchy, which has infinite variance.  If we truncate
to any lower=L and upper=U, then we get a finite variance.  So it
can have a profound effect in cases where the posteriors are broad. 

- Bob

ito4303: I see. I remove the constraint, and found it made the convergence considerably better. Thank you for your suggestion.

PR #49, opened 2016-03-03
BPA: Add Chapter 11

ito4303:
I translated models in Chapter 11 of BPA. Would you please review the code?


BPA/Ch.11/ipm.stan
```diff
@@ -0,0 +1,181 @@
...
+
+    if (lambda < 0) {
+      reject("lambda must not be negative; found lambda=", lambda);
+    } else if (n < 0.0) {
+      reject("n must not be negative; found n=", n);
+    } else {
+      return n * log(lambda) - lambda - lgamma(n + 1);
+    }
+    return negative_infinity();
+  }
```
bob-carpenter: Nice!  This is the best example we've seen yet of a user-defined density.

BPA/Ch.11/ipm.stan
```diff
@@ -0,0 +1,181 @@
...
+  for (t in 1:(nyears - 1))
+    rho[t] <- R[t] * f[t];
+}
+
+model {
+  // Priors
+  // Initial population sizes
+  // Constraints ensure truncated normal (> 0)
+  N1[1] ~ normal(100, 100);
+  Nad[1] ~ normal(100, 100);
```
bob-carpenter: I just want to make sure that 100 was the right mean here and it wasn't a typo.

BPA/Ch.11/ipm_hoopoe.stan
```diff
@@ -0,0 +1,222 @@
...
+  l_mphia ~ normal(0, 100);
+  l_mfec ~ normal(0, 100);
+  l_mim ~ normal(0, 100);
+  l_p ~ normal(0, 100);
+
+  // Improper flat priors are implicitly used on
+  // scale parameters, sig_phij, sig_phia, sig_fec and sig_im.
+
+  // Distribution of error terms
+  epsilon_phij ~ normal(0, sig_phij);
```
bob-carpenter: these can be more efficient statistically in non-centered form

```
a_raw ~ normal(0, 1);

a <- a_raw * sigma;
```

BPA/Ch.11/ipm_pred.stan
```diff
@@ -0,0 +1,228 @@
...
+    sad[t] <- mean_sad;
+    p[t] <- mean_p;
+    f[t] <- mean_fec;
+  }
+
+  // Total number of individuals
+  for (t in 1:nyears)
+    Ntot[t] <- Nad[t] + N1[t];
+
+  { // m-array
```
bob-carpenter: sometimes it can help readability if you break long blocks of code like this up into functions with informative names

bob-carpenter:
These look great.  I made some comments, but there's not really anything you need to change.  Are you making sure for these that you recover roughly the same estimates as the original code did with BUGS?

bob-carpenter:
Oh, just let me know if it's OK to merge or go ahead and merge it yourself (I don't know if you have that permission).

BPA/Ch.11/ipm.stan
```diff
@@ -0,0 +1,181 @@
...
+
+  mean_sjuv ~ uniform(0, 1);
+  mean_sad ~ uniform(0, 1);
+  mean_p ~ uniform(0, 1);
+  // Improper flat priors are implicitly used on sigma_y and mean_fec.
```
jgabry: Is it intentional that some uniform priors are explicitly written and others are left to be implied by constraints? I think it would be better to be consistent, but it's not a huge deal.

BPA/Ch.11/ipm_pred.stan
```diff
@@ -0,0 +1,228 @@
...
+  // Priors
+  // Initial population sizes
+  // Constraints ensure truncated normal (> 0)
+  N1[1] ~ normal(100, 100);
+  Nad[1] ~ normal(100, 100);
+
+  mean_sjuv ~ uniform(0, 1);
+  mean_sad ~ uniform(0, 1);
+  mean_p ~ uniform(0, 1);
+  // Improper flat priors are implicitly used on sigma_y and mean_fec.
```
jgabry: Same question here about why some uniform priors are written out explicitly and others and implicit.

BPA/Ch.11/ipm.stan
bob-carpenter: There's no way to write an improper uniform in Stan!  And the other ones are there for parallelism with the original models.

jgabry: Since parallelism with the originals is already broken by not being able to write an improper uniform I'd prefer to just drop all of the `uniform(0,1)` lines. It's kind of weird to be specifying the uniforms in two different ways within the same Stan program. Definitely not a big deal though. And it's really awesome to have all of these translations. Thanks again @ito4303 !!

bob-carpenter: Thanks for helping to analyze, Jonah.

I agree.  I'm OK either way, but prefer no explicit uniform()
distributions for consistency.  And I agree that it's great
Hiroki's building all these---thanks!  I think these models
are great.

It'd be nice to get some run-time comparisons vs. BUGS (in
terms of time divided by effective sample size(s)).

- Bob

BPA/Ch.11/ipm.stan
ito4303: In the original BUGS code, they are defined as `dnorm(100, 0.0001)I(0,)`.

BPA/Ch.11/ipm_pred.stan
ito4303: Thank you for the comments. Implicit prior declarations are now used for each model in this chapter if possible.

ito4303:
Thank you for reviewing. I always confirm that results are similar to true values (for simulated data) or the results of BUGS (for real data).

However, I usually do not compare running times. In case of the model in section 11.6 (ipm_hoopoe.stan), Stan took 97 seconds including compilation (4 chains in parallel, 12000 iterations/chain) while WinBUGS took 123 seconds (via Wine, no parallelization, 3 chains, 20000 iterations/chain) on my machine (MacBook Pro, 2.5 GHz Intel Core i7).

ito4303:
I don't think I have the permission to merge. Please merge the code if there is no problem.

BPA/Ch.11/ipm.stan
bob-carpenter: OK, we don't need the explicit truncation here in Stan
if the parameters N1 and Nad are declared with lower=0.

bob-carpenter:
That's probably way more iterations than you need for Stan.

You only need to run up to some reasonable effective sample
size for whatever inferences you need to make.

To compare times, you really want to measure time per
effective sample size.  And compute those effective sample
sizes with multiple chains using Stan's more conservative
n_eff estimator.  But no worries here --- you don't need to
do that for this issue, I was just curious.

PR #54, opened 2016-03-07
BPA: Add Chapter 12 (Section 12.2)

ito4303:
I add models in Chapter 12 (Section 12.2). Would you please review the code?


BPA/Ch.12/binmix.stan
```diff
@@ -0,0 +1,36 @@
...
+}
+
+parameters {
+  real<lower=0> lambda;    // Mean population size
+  real<lower=0,upper=1> p; // Detection probability
+}
+
+model {
+  // Priors
+  lambda ~ gamma(0.005, 0.005);
```
bob-carpenter: We strongly recommend against these diffuse gamma priors as they have bad statistical and computational properties (too much mass in the tail relative to the head of the distribution, primarily).  Andrew's written several papers about it that we reference in the manual.

What do you know about the population size ahead of time?

BPA/Ch.12/binmix_cov.stan
```diff
@@ -0,0 +1,58 @@
...
+
+transformed parameters {
+  vector[R] log_lambda;
+  matrix[R, T] logit_p;
+
+  log_lambda <- alpha0 + alpha1 * X;
+  for (i in 1:R)
+    for (j in 1:T)
+      logit_p[i, j] <- beta0 + beta1 * X[i];
+}
```
bob-carpenter: If you turn X into a real valued-vector, then you can do a lot of this with vector arithmtic more efficiently.

bob-carpenter:
These look great---I just had two comments (see above and in code).  I think the `gamma(0.005, 0.005)` should be dealt with, but the efficiency issue is optional.

BPA/Ch.12/binmix.stan
ito4303: The population size is unknown, but it should be not so large; an alternative prior proposed in the book is uniform(0, 10). So that now I use a Cauchy prior on the parameter.

bob-carpenter:
Great, and thanks for documenting the change for unsuspecting readers!

PR #56, opened 2016-03-12
BPA: Add Chapter13 (Section 13.3)

ito4303:
I add models in Section 13.3 of the BPA book. Would you please review the code?


BPA/Ch.13/site-occ.stan
```diff
@@ -0,0 +1,52 @@
...
+      1 ~ bernoulli(psi);
+      y[i] ~ bernoulli(p);
+    } else {
+      real lp[2];
+
+      lp[1] <- bernoulli_log(1, psi)          // Occurred
+        + bernoulli_log(0, p) * T;            // and not observed
+      lp[2] <- bernoulli_log(0, psi);         // Not occurred
+      increment_log_prob(log_sum_exp(lp[1], lp[2]));
+    }
```
bob-carpenter: There's a binary `log_sum_exp`, so this can just be:

```
increment_log_prob(log_sum_exp(bernoulli_log(1, psi) + bernoulli_log(0, p) * T,
                                                        bernoulli_log(0, psi)));
```

or slightly less efficiently with log_mix, 

```
increment_log_prob(log_mix(psi, T * bernoulli_log(0, p), 0));
```

You can still define local `lp1` and `lp2` if you think it's clearer than nested expressions.

BPA/Ch.13/site-occ_cov.stan
```diff
@@ -0,0 +1,73 @@
...
+}
+
+transformed parameters {
+  vector[R] logit_psi;            // Logit occupancy probability
+  matrix[R, T] logit_p;           // Logit detection probability
+
+  logit_psi <- alpha_occ + beta_occ * X;
+  for (j in 1:T)
+    logit_p[1:R, j] <- alpha_p + beta_p * X;
+}
```
bob-carpenter: You want to only compute `alpha_p + beta_p * X` once.  So this should probably be:

```
if (j > 0)
  logit_p[1:R, 1] <- alpha_p + beta_p * X;
for (j in 2:T)
  logit_p[1:R, j] <- logit_p[1:R, 1];
```

If you know `j > 0, you can remove that first conditional.  Or you could just assign the value to a temp and have a single loop.

It'd be more efficient for this step if logit_p were transposed, so you could assign to logit_p[j].  There's a lot of extra copying going on the way you have it (not that copies are that expensive).

bob-carpenter:
This is pretty much fine as is;  if you want to address the comments, feel free.  Just let me know when you want to merge.

ito4303:
Thank you for reviewing. For the latter issue, I have refined the code using `rep_matrix` function.

PR #58, opened 2016-03-18
BPA: Add Section 13.4-13.6

ito4303:
I add models in Section 13.4–13.6. Would you please review the code?


BPA/Ch.13/Dynocc.stan
```diff
@@ -0,0 +1,98 @@
...
+transformed parameters {
+  simplex[2] ps[2, nyear-1];     // Transition probability
+  simplex[nrep+1] po[2, nyear];  // Emission probability
+
+  for (t in 1:(nyear - 1)) {
+    ps[1, t, 1] <- phi[t];
+    ps[1, t, 2] <- 1.0 - phi[t];
+    ps[2, t, 1] <- gamma[t];
+    ps[2, t, 2] <- 1.0 - gamma[t];
+  }
```
bob-carpenter: Given that phi and gamma have `uniform(0,1)` priors, you could just define `ps` to be a parameter (with implicit uniform prior) directly.  

Also, if you need more arithmetic stability, you could define `log_po` instead of `po`, then `gam` would be defined by `log_sum_exp(...)` instead of `log(sum(...))` and you'd need some other logs.  Definitely not worth doing here if the model works, but something to keep in mind if the `exp(binomial_log(...))` expression starts underflowing.

BPA/Ch.13/owls_ms1.stan
```diff
@@ -0,0 +1,77 @@
...
+  // Likelihood
+  for (s in 1:R) {
+    vector[3] lp;
+
+    for (k in 1:3) {
+      lp[k] <- categorical_log(k, phi[s]);
+      for (t in 1:T) {
+        if (y[s, t])
+          lp[k] <- lp[k] + categorical_log(y[s, t], p[k, t]);
+      } //t
```
bob-carpenter: It's up to you here, but I find the comments on close-loop brackets to be distracting because I'm used to more following the shapes of programs where the close brackets are distinctive in part for being on their own lines.

bob-carpenter:
Thanks!  These look great.  I'm glad to see they're still matching the JAGS output.  Let me know when you want this merged.

BPA/Ch.13/Dynocc.stan
ito4303: I tried to transform `po` into `log_po`, but this attempt resulted in sampling error with initial values maybe because `log_po` contained many negative infinities.

ito4303:
Thank you for reviewing. Please merge the code if there are no more problems.

PR #59, opened 2016-03-26
BPA: Add Section 12.3 (final)

ito4303:
I add models in Section 12.3 of BPA. This is the final section to complete the translation. Would you please review the code?


bob-carpenter:
This all looks fantastic. Thanks so much.  I'll merge it.

PR #60, opened 2016-03-31
Gam

yannickspill:
This is an example implementation of Generalized Additive Models, as described by S. Wood in his 2006 book of the same name, and implemented in the R package `mgcv`. This pull request follows from a discussion in the google groups, topic `Positivity and centering constraints in stan`. A few notes:
- This implementation of GAM uses p-splines, and corresponds to the formula `y~s(x,bs="ps")`. I took some freedom to define the prior but I think it is consistent with what is described in the book.
- It is possible to use other distributions by changing a line in the model.
- This implementation centers the spline base and adds an intercept. That way, other terms can be added to the formula by just appending columns to the design matrix.
- It is very slow to whatever `gam` from `mgcv` achieves, and I don't know how to accelerate it.
- I am working on a sparse version which will center the parameter vector instead, which would allow a large spline base.
- other formulations, including SCAM (e.g. GAM with monotonicity constraints) are available, but let's start with this to keep it simple.


misc/gam/gam_one_centered_design.stan
```diff
@@ -0,0 +1,177 @@
...
+  P <- difference_op(x, K, splinedegree(), difforder());
+}
+parameters {
+  vector[K] beta; //parameter vector for intercept + centered spline
+  real<lower=0> alpha; //precision of negative binomial
+  real<lower=0> lambda; //regularization parameter
+}
+model {
+  //negative binomial GAM
+  y ~ neg_binomial_2(exp(X * beta), alpha);
```
bob-carpenter: You should use

```
y ~ neg_binomial_2_log(X* beta, alpha);
```

here.  It's the same function, but we do the algebra on the inside more robustly and efficiently than just applying `exp(...)`.

misc/gam/gam_one_centered_design.stan
```diff
@@ -0,0 +1,177 @@
...
+  //P-spline prior on the 2nd order differences (K-2 params)
+  //warning on jacobian can be ignored
+  //see GAM, Wood (2006), section 4.8.2 (p.187)
```
bob-carpenter: I'd prefer to see these notes in model comments at the top rather than inline.  The problem with putting anything other than a simple end-of-line comment inline is that it makes the code hard to read as code.

misc/gam/gam_one_centered_design.stan
```diff
@@ -0,0 +1,177 @@
...
+  P*beta ~ normal(0, 1./(alpha*lambda));
+}
```
bob-carpenter: Don't need those `1.` things --- it looks like a holdover from another language.

And we'd prefer spaces between operators like `*` and definitely between `+` and `-`.

misc/gam/gam_one_centered_design.stan
```diff
@@ -0,0 +1,177 @@
...
+  vector[N] pred; //spline interpolant
+  real edf; //effective degrees of freedom
+  vector[K] edfvec;
+  designmat <- X;
+  weighted <- X .* rep_matrix(beta', rows(x));
+  pred <- exp(X * beta);
+  {
+    matrix[K,K] XtX;
+    XtX <- crossprod(X);
+    edfvec <- diagonal(inverse_spd(XtX+lambda*crossprod(P)) * XtX);
```
bob-carpenter: You don't ever want to do inverses if you can avoid them.  So rather than inv(a) * b, you want `a \ b`.  But then I don't think we have an spd-specific version (symmetric pos definite).

And we'd also prefer spaces between `N,K` as in `N, K`.

misc/gam/gam_one_centered_design.stan
```diff
@@ -0,0 +1,177 @@
...
+  }
+  
+  //apply centering constraint on X to a matrix D: D -> DZ
+  // where Z is such that X^T 1 = QR with Q=[u:Z]
+  // if D is LxK, and X is NxK, returns a Lx(K-1) matrix
+  matrix center(matrix X, matrix D) {
+    vector[cols(X)] u;
+    u <- centering_constraint(X);
+    return D[,2:] - (2*D*u)*u[2:]';
+  }
```
bob-carpenter: I'd much rather see doxygen/Javadoc-style doc here, as in:

```
/**
  * Return ...explanation of what gets returned and the math...
  * @param X ... short description of X ...
  * @param D ... short description of D ...
  * @return Centered version of D.
  */

LIke in C++, we like to reserve the `//` for simple end-of-line comments and reserve `/* ... */` for doc comments.

File: misc/gam/gam_one_centered_design.stan [hunk 6]
```diff
@@ -0,0 +1,177 @@
+////
+// Generalized Additive Model : y ~ s(X)
+// using cubic splines with negative binomial and log-link
+// The spline design matrix is centered and augmented with an intercept
+// Author: Yannick G Spill
+// References:
+// - Generalized Additive Models: an introduction with R, Simon N. Wood, CRC press (2006)
+// - Eilers and Marx, Stat Sci, 1996
+// This work is licensed under a Creative Commons Attribution-ShareAlike 4.0 International License
```
bob-carpenter: That's a doc license, not a code license.  Would you mind 3-clause BSD (aka "new BSD"?)

And you need to indicate who the copyright holder is --- it's usually either you or your employer depending on the arrangement you have with them (many universities, including Columiba, assert copyright ownership of code written by their employees).

And thanks for including the references.

bob-carpenter:
Thanks a lot for the contribution.  I made a bunch of comments inline on the code.  I don't see easy speed improvements unless the matrix operations could be cleaned up (not really my specialty, but someone else may want to jump in).

bob-carpenter:
@yannickspill  Just let me know if you want to change anything before merging this.

yannickspill:
@bob-carpenter I'm a bit swamped too right now, couldn't contact my boss about the licensing either. If you don't mind we can give it a week or two?

bob-carpenter:
No rush whatsoever.  Thanks.

yannickspill:
Bob
Can you suggest a licence that would be free for academic use, but that would not allow companies to use it for profit?

bob-carpenter:
None of the standard open-source licenses have any
terms relating to profit or companies vs. academia.

Nor would we accept contributions made under such
a restrictive license.

See:  https://opensource.org

and in particular:  https://opensource.org/osd

- Bob

yannickspill:
Hi @bob-carpenter I will check with the hierarchy to see what they say of the 3-clause bsd. What about lgpl or gpl? is that fine with you too?

bob-carpenter:
For example models, GPL or LGPL is OK, though I'd
prefer BSD.

But not for any code that went into our core libraries
in repos stan-dev/stan or stan-dev/math.

- Bob

yannickspill:
Hi @bob-carpenter sorry for the long, long delay. Let's go for the 3-clause BSD.

PR #67, opened 2016-06-16
curse of dims first draft

bob-carpenter:
I wrote a "curse of dimensionality" case study in example-models/knitr/curse-dims.  If someone would like to review it before I post it on our web site, please do.  Otherwise, I'll just merge it and put it up in a week or so.


betanalpha:
Section 1: I’d add a third section discussion how almost all pairs of points are “far away”
from each other in high-dimensions.  Take two points that are translated from each other
by 1 unit in each direction — the distance between them grows as sqrt(d) so as the
dimension grows the points appear further from each other, with more and more volume
filling up the space between them.  Per https://twitter.com/bschoelkopf/status/503554842829549568,
“a high-dimensional space is a lonely place”.

Section 2: The first square corner is not typeset correctly.  I’d also add a graph showing the
number of corners vs dimension.  I don’t think the topological discussion is particularly useful
here.  Finally you might want to be careful with the nomenclature here — there’s lots of volume
in the corners, which implies that the mass of the uniform distribution also accumulates in the
corners, but not _all_ probability mass does that.  This is important because later you show
probability concentrating in a sphere and not in the corners because you have a balance
between the concentration of uniform mass in the corners and some density near the mode
in the center.  To avoid this ambiguity I like reserving “volume” for uniform mass (or really 
Lebesgue mass) and “mass” for the probability mass of a particular target distribution,
but I’ll leave the vocabulary to you.

Section 4: May be worth noting that the draws are not only moving away from the mode but
also clustering tighter and tighter relative to the distance.  In higher dimensions this concentration
becomes so strong that the draws are essentially all at the same radius.

Section 5: This is a really nice opportunity to mention an important phenomenon — in high
dimensions no individual in a population looks like the average!  In other words, there’s no
single “representative” individual of the entire population.  Funnily enough in particle physics
they make this mistake all the time, leading to “Asimov data sets” 
(https://en.wikipedia.org/wiki/Franchise_(short_story)) and horrendously biased representations.

I’d also mention that the trend of all probability mass being drawn into a more and more singular
typical set is called “concentration of measure” and is fundamentally a probabilistic notion.  Of
course I’d argue that information theory is just applied measure theory, but I wouldn’t want to
step on any computer scientists toes.

Section 6:  I’d move this to Section 4.  Then you can plot the mean +/- std dev to complement
the empirical distribution of the draws.

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+grows, most of the points in the hypercube lie outside the inscribed
+hypersphere.
+
+Suppose we have an $N$-dimensional hypercube, with unit-length sides
+centered around the origin $\mathbf{0} = (0, \ldots, 0)$.  The
+hypercube will have $2^N$ corners at the points $\left( \pm
+\frac{1}{2}, \ldots, \pm \frac{1}{2} \right)$.  Because its sides are
+length 1, it will have also have unit hypervolume, because $1^N = 1$.
+
+If $N=1$, the hypercube is a line from $\left( -\frac{1}{2}, \right)$
```
syclik: this should be: ```$\left( -\frac{1}{2} )$```

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+to $\left( -\frac{1}{2}, \right)$ of unit length (i.e., the length is
```
syclik: this should be: ` $\left( \frac{1}{2} \right)$ \`
(notice: sign change and removal of comma)

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+1).  If $N=2$, the hypercube is a square of unit area with opposite
+corners at $($\left( -\frac{1}{2}, -\frac{1}{2} \right)$ and $\left(
```
syclik: Remove the first `$(` in this bit of latex: `$($\left( -\frac{1}{2}, -\frac{1}{2} \right)$`

knitr/curse-dims/curse-dims.Rmd
syclik: drop the parens here and for the upper limit of this in 1 dimension to be consistent with the later section.

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+\frac{1}{2}, \frac{1}{2} \right)$.  With $N=3$, we have a cube of unit
+volume, with opposite corners at $\left( -\frac{1}{2}, -\frac{1}{2},
+-\frac{1}{2} \right)$ and $\left( \frac{1}{2}, \frac{1}{2},
+\frac{1}{2} \right)$, and unit volume.  And so on up the dimensions.
+
+Now consider the biggest hypersphere you can inscribe in the
+hypercube.  It will be centered at the origin and have a radius of
+$\frac{1}{2}$ so that it extends to the sides of the hypercube.  A
+point $y$ is within this hypersphere if the distance to the origin is
+less than or equal to the radius, or in symbols, if $||y|| \leq
```
syclik: change to an "open ball" instead of "closed ball" to match the simulation code and everything below. (We could use a closed ball, but then we'd have to change all the simulation code.)

So, the places to change:
- here: `$||y|| < \frac{1}{2}$`
- the text in the next sentence
- talk about using an open ball and in parens, talk about how we could've worked with a closed ball.

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+middle <- rep(NA, D);
+upper <- rep(NA, D);
+lower_ll <- rep(NA, D);
+middle_ll <- rep(NA, D);
+upper_ll <- rep(NA, D);
+
+mean_ll <- rep(NA, D);
+max_ll <- rep(NA, D);
+for (k in 1:D) {
+  d <- as.integer(sqrt(2)^(k - 1));
```
syclik: I don't understand this. For k = 1:D, it produces:
1, 1, 2, 2, 4, 5, 8, 11, 16, 22, 32, 45, 64, 90, 128, 181, 256.

The real question is why this is repeating 1 and 2 twice?

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+plot2;
+```
+
+## 5. Typical Sets
+
+Roughly speaking, the typical set is where draws from a given
+distribution tend to lie.  That is, the typical set encompasses almost
+all of the probability mass for a distribution (sets of outcomes for
+discrete distributions and volumes for continuous ones).  What we will
+see in this section is that the typical set is usually nowhere nearthe
```
syclik: change "nearthe" to "near the"

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+namely 
+
+\[
+\mathsf{Binomial}(y \, | \, N, \theta) 
+= \binom{N}{y} \, \theta^y \, (1 - \theta)^{(N - y)},
+\]
+
+where the binomial coefficient that normalizes the distribution is
+defined as the number of binary sequences of length $N$ that contain
+exactly $y$ 1 values,
```
syclik: is it weird to make "1 values" a hyphenated compound noun? I had trouble parsing this the first time around.

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+
+That is, it's the set we need to compute expectations.  The fact that
+typical sets exist is why we are able to compute expectations using
+Monte Carlo methods.
+
+We saw in the first plot in section 4 (draws are nowhere near the
+mode) that if we have $D$ i.i.d. draws from a unit normal, that the
+typical set was well bounded away from the posterior mode, which is at
+the origin.  There's no built-in R function to compute the quantiles
+directly, but the next section shows that the distaince from the mode
```
syclik: spelling: "distaince" should be "distance"

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+1$ and unknown location $\mu$.  Then the log density is
+
+\[
+\log p(y \, | \, \mu, 1) 
+\ \propto \
+\sum_{n=1}^N \log \mathsf{Normal}(y_n \, | \, \mu, 1)
+\ \propto \
+-\frac{1}{2} \sum_{n=1}^N \left( y_n - \mu \right)^2
+\ \propto \
+-\frac{1}{2} \sum_{n=1}^N \mathrm{d}(y_n, \mu).
```
syclik: need to square the `d(y_n, \mu)` term:
-\frac{1}{2} \sum_{n=1}^N \mathrm{d}(y_n, \mu)^2.

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+\]
+
+The maximum likelihood estimate for the location $\mu$ is just the
+value that maximizes the likelihood function,
+
+\[
+\mu^* 
+\ = \
+\mathrm{argmax}_{\mu} \ p(y \, | \, \mu),
```
syclik: remove comma at the end of the line

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+\ = \
+\mathrm{argmax}_{\mu} \ \log p(y \, | \, \mu),
```
syclik: remove comma at the end of the line

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+
+## Exercises
+
+1. Define an R function for the Euclidean distance between two vectors.
+What happens if you use R's matrix multiplication, such as `u %*% v`?
+
+1.  Given $y_n \in \mathbb{R}$ for $n \in 1{:}N$, the maximum
+likelihood estimate of $\mu$ for the model 
+\[
+p(y | \mu) = \prod_{n=1}^N \mathsf{Normal}(y_n \, | \, \mu, 1)
```
syclik: add additional horizontal space around `|`:
`p(y \, | \, \mu) = ...`

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+\]
+
```
syclik: I think this blank newline is causing the output to think it should start renumbering. Remove it and see if it works?

syclik: On second thought, I see this in a few places in the exercises; I'm not sure I know how to handle the numbering well enough.

bob-carpenter: It renders them in order.  I could've also used 0.  Basically,
it ignores the value of the number.

bob-carpenter: It renders OK in the numbering for me.

knitr/curse-dims/curse-dims.Rmd
```diff
@@ -0,0 +1,770 @@
...
+value that maximizes the likelihood function,
+
+\[
+\mu^* 
+\ = \
+\mathrm{argmax}_{\mu} \ p(y \, | \, \mu),
+\ = \
+\mathrm{argmax}_{\mu} \ \log p(y \, | \, \mu),
+\ = \
+\mathrm{argmax}_{\mu} - \sum_{n=1}^N \mathrm{d}(y_n, \mu)^2
```
bob-carpenter: Right.  The equality holds, but it's clearer with the 1/2 still
in the penultimate line.  I then explicitly mention dropping it.

knitr/curse-dims/curse-dims.Rmd
syclik: Sorry -- should have been clearer. It wasn't the use of the number 1 repeatedly that threw me off. In the generated html file, it starts renumbering back to 1 after the second exercise. I assumed it was due to the extra whitespace signaling to markdown that the numbering was done.

knitr/curse-dims/curse-dims.Rmd
bob-carpenter: It's the same problem as the autodiff paper.  That operation
isn't dense in the progression I have multiplying by sqrt(2)
rather than 2 to keep everything even on the log2 scale.

I'll just hardcode the sequence and sizes here as
it'll be less confusing.

knitr/curse-dims/curse-dims.Rmd
bob-carpenter: Just rephrased that.  And I did all the other minor typo things
you suggested.  Thanks for reading so carefully.

knitr/curse-dims/curse-dims.Rmd
bob-carpenter: fixed.

knitr/curse-dims/curse-dims.Rmd
bob-carpenter: fixed, not sure how that even parsed.

knitr/curse-dims/curse-dims.Rmd
bob-carpenter: If that was a problem, it's fixed now and renders properly, at
least for me.

bob-carpenter:
Thanks.  That's really helpful and I'll try to rewrite.  I was
reading more McKay on the train and came across a really nice
illustration of concentration of measure that I'm going to
include.  It drew the same graph you drew of density vs.
volume, with the typical set shell being the product, except he
did it for a binomial.  It illustrates that tighter banding
you suggest very neatly.  That book's packed with insight.

Of course information theory is applied measure theory!  It
gives you a more coding-oriented perspective, which is really
nice to apply to noisy channels and communication.  That's why
it's mainly EE people, not CS people who do it.  The same crowd
that brought you Kalman filtering and "uncertainty quantification".

And I was also thinking about what you mean that no point looks
average.  I was thinking why the mode isn't representative, but
then neither is any other single sequence.  Writing this stuff
up is great for making you think about it in a way that reading
doesn't.

PR #68, opened 2016-06-17
Arm ch12 model updates

ericnovik:
Fixed Stan files in ARM ch12, put in proper priors and fixed constraints on scale parameters.

@jgabry, please take a look.


ARM/Ch.12/radon_group_chr.stan
```diff
@@ -1,17 +1,17 @@
 data {
-  int<lower=0> N; 
+  int<lower=0> N;
   int<lower=1,upper=85> county[N];
```
jgabry: @ericnovik not a big deal, but might as well use `J` here too instead of `85` like you do in the others

ARM/Ch.12/radon_intercept_chr.stan
```diff
@@ -1,14 +1,14 @@
 data {
-  int<lower=0> N; 
+  int<lower=0> N;
   int<lower=1,upper=85> county[N];
   vector[N] y;
-} 
+}
```
jgabry: `J` here instead of `85` too I guess

ericnovik: @jgabry good catch, thanks; will correct

ericnovik:
@jgabry I fixed the hard coded value (85 -> J)

PR #76, opened 2016-10-28
fixed name of model file in R script

dallascard:
One of the Binomial models had a misspecified file name in the R script.


jgabry:
Thanks!

PR #79, opened 2016-12-16
BPA/Ch.05 & Ch.06 update

ito4303:
Would you review the code?

### DIFF
--- a/BPA/Ch.06/Mh.stan
+++ b/BPA/Ch.06/Mh.stan
@@ -7,27 +7,27 @@ data {
 transformed data {
   int<lower=0> C;               // Size of observed data set
 
-  C <- 0;
+  C = 0;
   for (i in 1:M) {
     if (y[i] > 0)
-      C <- C + 1;
+      C = C + 1;
   }
 }
 
 parameters {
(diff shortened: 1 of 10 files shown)


BPA/Ch.06/Mh.stan
```diff
@@ -7,27 +7,27 @@ data {
...
 }
 
 parameters {
-  real<lower=0,upper=1> omega;  // Inclusion probability
-  real<lower=0,upper=1> mean_p; // Detection probability
-  real<lower=0> sigma;
-  real<lower=-16,upper=16> eps[M];      // Individual random effects
+  real<lower=0,upper=1> omega;      // Inclusion probability
+  real<lower=0,upper=1> mean_p;     // Mean detection probability
+  real<lower=0,upper=5> sigma;
```
bob-carpenter: Are these new upper bounds to make the models match what is in the book?  If so, that's fine, but in general we don't recommend these hard upper bounds on scale parameters---we use softer bounds with no hard constraints on the upper bound.

ito4303: Yes, they are defined to match the model in the BPA book. I know that they are not recommended in Stan.

PR #80, opened 2016-12-17
BPA/Ch.06 update

ito4303:
I have changed priors of scale parameters into improper uniform or weakly informative ones for efficiency, irrespective of the original models in the book. Would you review the code?

### DIFF
--- a/BPA/Ch.06/6.2.4_Model_Mh.R
+++ b/BPA/Ch.06/6.2.4_Model_Mh.R
@@ -21,8 +21,8 @@ params <- c("N", "mean_p", "sigma", "omega")
 
 ## MCMC settings
 ni <- 15000
-nt <- 5
-nb <- 10000
+nt <- 10
+nb <- 5000
 nc <- 4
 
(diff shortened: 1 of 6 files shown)


bob-carpenter:
We've talked about this for our BUGS model translations and decided to keep a literal translation of the model (that is, one that has the same log density).  Then if we have what we think is a better approach, include that version, too.  Would it be hard to follow that pattern here?

Also, if you want to replace a `uniform(0, 3)` prior, you can use a `normal(1.5, 0.75)`, which puts 95% of the mass within `(0, 3)` and allows a thin tail to the right (assuming the `<lower=0>` declaration stays.  Or you can do the same with fatter tails---put most of the mass inside the original `(0, 3)` bound.   Or you can use a lognormal with appropriate location and scale and that will avoid values very close to zero.

ito4303:
Thank you for your advice. I changed the priors to match the original model in the book again. In addition, I added comments to indicate how to use weakly informative priors.

PR #82, opened 2016-12-21
BPA/Ch.07 update

ito4303:
I updated the models in Ch.07 of BPA. Would you review the code?

### DIFF
--- a/BPA/Ch.07/cjs_cov_raneff.stan
+++ b/BPA/Ch.07/cjs_cov_raneff.stan
@@ -26,15 +26,18 @@ functions {
     for (i in 1:nind) {
       chi[i, n_occasions] = 1.0;
       for (t in 1:(n_occasions - 1)) {
+        // Compoud declaration was enabled in Stan 2.13
+        int t_curr = n_occasions - t;
+        int t_next = t_curr + 1;
+        /*
         int t_curr;
         int t_next;
 
         t_curr = n_occasions - t;
         t_next = t_curr + 1;
+        */
         chi[i, t_curr] = (1 - phi[i, t_curr])
-                       + phi[i, t_curr]
-                       * (1 - p[i, t_next - 1])
-                       * chi[i, t_next];
+                        + phi[i, t_curr] * (1 - p[i, t_next - 1]) * chi[i, t_next];
       }
     }
     return chi;
@@ -45,14 +48,17 @@ data {
   int<lower=0> nind;            // Number of individuals
   int<lower=2> n_occasions;     // Number of capture occasions
   int<lower=0,upper=1> y[nind, n_occasions];    // Capture-history
-  real x[n_occasions - 1];      // Covariate
+  vector[n_occasions - 1] x;    // Covariate
 }
 
 transformed data {
+  int n_occ_minus_1 = n_occasions - 1;
+  //  int n_occ_minus_1;
   int<lower=0,upper=n_occasions> first[nind];
   int<lower=0,upper=n_occasions> last[nind];
   vector<lower=0,upper=nind>[n_occasions] n_captured;
 
+  //  n_occ_minus_1 = n_occasions - 1;
   for (i in 1:nind)
     first[i] = first_capture(y[i]);
   for (i in 1:nind)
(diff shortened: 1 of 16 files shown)


bob-carpenter:
I just realized you could simplify some of this code.

```
psi = inv_logit(...);
z ~ bernoulli(psi);
```

is equivalent to

```
logit_psi = ...;
z ~ bernoulli_logit(logit_psi);
```

We have Bernoulli that takes a parameter on the logit scale, so you don't need an explicit `inv_logit`.  I should update the manual!

BPA/Ch.07/cjs_mnl_ran.stan
```diff
@@ -24,47 +24,50 @@ transformed parameters {
...
   simplex[n_occasions] pr[n_occasions - 1];
 
-  mu <- logit(mean_phi);
+  mu = logit(mean_phi);
 
   // Constraints
   for (t in 1:(n_occasions - 1)) {
-    phi[t] <- inv_logit(mu + epsilon[t]);
-    p[t] <- mean_p;
+    phi[t] = inv_logit(mu + epsilon[t]);
```
bob-carpenter: We have a lot more functions vectorized, so if you can't work on the logit scale and just define

```
logit_phi = mu + epsilon;
```

then you can do this:

```
phi = inv_logit(mu + epsilon);
```

though you may need to slice down to `1:(n_occassions - 1)` if that isn't the size of `mu`.

BPA/Ch.07/cjs_mnl_ran.stan
```diff
@@ -24,47 +24,50 @@ transformed parameters {
...
     for (j in 1:(t - 1))
-      pr[t][j] <- 0;
-  } // t
+      pr[t][j] = 0;
+  }
 
   // Last column: probability of non-recapture
   for (t in 1:(n_occasions - 1))
-    pr[t][n_occasions] <- 1 - sum(head(pr[t], n_occasions - 1));
+    pr[t][n_occasions] = 1 - sum(pr[t][1:(n_occasions - 1)]);
```
bob-carpenter: You can have this be `pr[t, 1:(n_occasions - 1)]`.

Given how often you use this `n_occasions - 1` calculation, you might want to make a transformed data item and then just reuse it.

BPA/Ch.07/cjs_mnl_ran.stan
```diff
@@ -77,25 +80,25 @@ generated quantities {
   real fit;
   real fit_new;
 
-  sigma2 <- square(sigma);
+  sigma2 = square(sigma);
 
   // Temporal variance on real scale
-  sigma2_real <- sigma2 * square(mean_phi) * square(1 - mean_phi);
+  sigma2_real = sigma2 * square(mean_phi) * square(1 - mean_phi);
```
bob-carpenter: It's more efficient to reduce the overall number of operations that go onto the expression graph, so by that principle, `square(a * b)` is more economical than `square(a) * square(b)`.  The first only has two multiplications, whereas the latter has three.  It's no big deal, especially in generated quantities where there's no autodiff overhead.   So if you think the way you have it is clearer, please leave it that way.

ito4303:
Thank you for reviewing the code. I'd like to repost the updated models after checking in Stan 2.13 or later.

ito4303:
I have updated the models and checked using Stan 2.14. For consistency, ``bernoulli()`` is left unchanged.

BPA/Ch.07/cjs_cov_raneff.stan
```diff
@@ -24,17 +24,17 @@ functions {
...
         int t_curr;
         int t_next;
 
-        t_curr <- n_occasions - t;
-        t_next <- t_curr + 1;
-        chi[i, t_curr] <- (1 - phi[i, t_curr]) +
-                          phi[i, t_curr] *
-                          (1 - p[i, t_next - 1]) *
-                          chi[i, t_next];
+        t_curr = n_occasions - t;
```
ito4303: I adopted compound declarations and commented out previous notations.

bob-carpenter:
Let me know if there's anything else you want me to look at.  Otherwise, you can just go ahead and merge (or I can if you'd rather).

ito4303:
Thank you for reviewing. I have no more changes in this chapter. Please merge the updtes if they look good.

PR #87, opened 2017-01-15
BPA/Ch.10 update

ito4303:
I updated the models in BPA/Ch.10 with 2.13 syntax, alongside putting common routines (calculating likelihoods) into functions. Would you review the code?

### DIFF
--- a/BPA/Ch.10/js_ms.stan
+++ b/BPA/Ch.10/js_ms.stan
@@ -10,6 +10,29 @@
 // 2 not seen
 //--------------------------------------
 
+functions {
+  /**
+   * Returns delta, where
+   * delta[n] = gamma[n] * PROD_{m < n} (1 - gamma[m])
+   * (Thanks to Dr. Carpenter)
+   *
+   * @param gamma Vector of probability sequence
+   *
+   * @return Vector of complementary probability sequence
+   */
+  vector seq_cprob(vector gamma) {
+    int N = rows(gamma);
+    vector[N] log_cprob;
+    real log_residual_prob = 0;
+
+    for (n in 1:N) {
+      log_cprob[n] = log(gamma[n]) + log_residual_prob;
+      log_residual_prob = log_residual_prob + log(1 - gamma[n]);
+    }
+    return exp(log_cprob);
+  }
+}
+
 data {
   int<lower=0> M;                               // Augmented sample size
   int<lower=0> n_occasions;                     // Number of capture occasions
@@ -44,7 +67,7 @@ transformed parameters {
     for (t in 1:n_occ_minus_1) {
       ps[1, i, t, 1] = 1.0 - gamma[t];
       ps[1, i, t, 2] = gamma[t];
-      ps[1, i, t, 3] =  0.0;
+      ps[1, i, t, 3] = 0.0;
       ps[2, i, t, 1] = 0.0;
       ps[2, i, t, 2] = phi[t];
       ps[2, i, t, 3] = 1 - phi[t];
@@ -111,17 +134,12 @@ generated quantities {
 
   // Calculate derived population parameters
   {
-    vector[n_occ_minus_1] qgamma;
-    vector[n_occ_minus_1] cprob;
+    vector[n_occ_minus_1] cprob  = seq_cprob(gamma[:n_occ_minus_1]);
     int al[M, n_occ_minus_1];
     int d[M, n_occ_minus_1];
     int alive[M];
     int w[M];
 
-    qgamma = 1.0 - gamma;
-    cprob[1] = gamma[1];
-    for (t in 2:n_occ_minus_1)
-      cprob[t] = gamma[t] * prod(qgamma[1:(t - 1)]);
     psi = sum(cprob);
     b = cprob / psi;
 
@@ -129,7 +147,7 @@ generated quantities {
       for (t in 2:n_occasions)
         al[i, t - 1] = (z[i, t] == 2);
       for (t in 1:n_occ_minus_1)
-        d[i, t] = (z[i, t] - al[i, t] == 0);
+        d[i, t] = (z[i, t] == al[i, t]);
       alive[i] = sum(al[i]);
     }
 
@@ -138,7 +156,7 @@ generated quantities {
       B[t] = sum(d[, t]);
     }
     for (i in 1:M)
-      w[i] = (alive[i] != 0);
+      w[i] = 1 - !alive[i];
     Nsuper = sum(w);
   }
 }
--- a/BPA/Ch.10/js_ms.stan
+++ b/BPA/Ch.10/js_ms.stan
@@ -16,47 +16,51 @@ data {
   int<lower=1,upper=2> y[M, n_occasions];       // Augmented capture-history
 }
 
+transformed data {
+  int n_occ_minus_1 = n_occasions - 1;
+}
+
 parameters {
-  vector<lower=0,upper=1>[n_occasions-1] gamma; // Removal entry probabilities
+  vector<lower=0,upper=1>[n_occ_minus_1] gamma; // Removal entry probabilities
   real<lower=0,upper=1> mean_phi;               // Mean survival
   real<lower=0,upper=1> mean_p;                 // Mean capture
 }
 
 transformed parameters {
-  vector<lower=0,upper=1>[n_occasions-1] phi; // Survival probability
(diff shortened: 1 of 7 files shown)


BPA/Ch.10/js_ms.stan
```diff
@@ -16,133 +16,129 @@ data {
...
+    }
+
+    for (t in 1:n_occ_minus_1) {
+      N[t] = sum(al[, t]);
+      B[t] = sum(d[, t]);
+    }
     for (i in 1:M)
-      w[i] <- (alive[i] != 0);
-    Nsuper <- sum(w);
+      w[i] = (alive[i] != 0);
```
bob-carpenter: This can just be `!alive[i]` --- you usually don't want to be comparing booleans to 1 and 0.  Sorry I didn't catch a lot of this the first review.  There's always something to make better with code.

BPA/Ch.10/js_rest_occ.stan
```diff
@@ -186,45 +204,43 @@ generated quantities {
...
     real mu2;
 
-    z[i, 1] <- bernoulli_rng(gamma[1]);
+    z[i, 1] = bernoulli_rng(gamma[1]);
     for (t in 2:n_occasions) {
-      q[t - 1] <- 1 - z[i, t - 1];
-      mu2 <- phi[i, t - 1] * z[i, t - 1] +
-             gamma[t] * prod(q[1:(t - 1)]);
-      z[i, t] <- bernoulli_rng(mu2);
+      q[t - 1] = 1 - z[i, t - 1];
```
bob-carpenter: This looks like that same kind of loop again, so maybe the function could be reused?  Or something like it to avoid the `N^2` operations?

BPA/Ch.10/js_rest_occ.stan
```diff
@@ -186,45 +204,43 @@ generated quantities {
...
-      qgamma[t] <- 1 - gamma[t];
-    cprob[1] <- gamma[1];
+    qgamma = 1 - gamma[1:(n_occasions - 1)];
+    cprob[1] = gamma[1];
     for (t in 2:n_occasions)
-      cprob[t] <- gamma[t] * prod(qgamma[1:(t - 1)]);
-    psi <- sum(cprob);
-    for (t in 1:n_occasions)
-      b[t] <- cprob[t] / psi;
+      cprob[t] = gamma[t] * prod(qgamma[1:(t - 1)]);
```
bob-carpenter: I'll let you find the rest!

BPA/Ch.10/js_super.stan
```diff
@@ -222,24 +241,22 @@ generated quantities {
     int Nind[M];
     int Nalive[M];
 
+    for (i in 1:M)
+      for (t in 1:n_occasions)
+        u[i, t] = z[i, t] * w[i];
```
bob-carpenter: This loop can be replaced with an elementwise product, which is more concise and also more efficient because the expression graph is smaller:

```
u[i] = z[i] .* w[i];
```

if `u` is a matrix, `z` is a matrix, and `w` is a vector.

BPA/Ch.10/js_super_indran.stan
```diff
@@ -104,98 +182,40 @@ transformed parameters {
...
   // Priors
-  // An improper flat prior (0,+infinity) is implicitly used on sigma.
-  mean_phi ~ uniform(0, 1);
-  mean_p ~ uniform(0, 1);
+  // Uniform priors are implicitly defined.
+  //  mean_phi ~ uniform(0, 1);
+  //  mean_p ~ uniform(0, 1);
+  //  psi ~ uniform(0, 1);
+  //  sigma ~ uniform(0, 5);
+  // In case of using a weakly informative prior
```
bob-carpenter: The above comments are fine, because they are bout what the code does.  The following comment about weakly informative priors opens a can of worms.  When you start making comments like this, where do you stop?  Probably better not to include this in the code. Maybe you could write a paper or monograph on how to code all these models in Stan and provide advice there?

BPA/Ch.10/js_super_indran.stan
```diff
@@ -231,24 +250,22 @@ generated quantities {
     int Nind[M];
     int Nalive[M];
 
+    for (i in 1:M)
+      for (t in 1:n_occasions)
+        u[i, t] = z[i, t] * w[i];
```
bob-carpenter: elementwise product

BPA/Ch.10/js_super_indran.stan
```diff
@@ -231,24 +250,22 @@ generated quantities {
...
-        u[i, t] <- z[i, t] * w[i];
-      }
-    }
-    for (i in 1:M) {
-      recruit[i, 1] <- u[i, 1];
+      recruit[i, 1] = u[i, 1];
       for (t in 2:n_occasions)
-        recruit[i, t] <- (1 - u[i, t - 1]) * u[i, t];
-    } //i
+        recruit[i, t] = (1 - u[i, t - 1]) * u[i, t];
```
bob-carpenter: These can also be done with elementwise products and slicing as

```
recruit[i, 1] = u[i, 1];
recruit[i, 2:n_occassions] = (1 - u[i, 1:(n_occassions - 1)]) .* u[i, 2:n_occasions];
```

You'll want to double check I got my algebra/indexing right if you want to make these changes.

BPA/Ch.10/js_tempran.stan
```diff
@@ -80,103 +149,52 @@ transformed data {
...
   // Priors
-  mean_phi ~ uniform(0, 1);
-  mean_p ~ uniform(0, 1);
-  gamma ~ uniform(0, 1);
+  // Uniform priors are implicitly defined.
+  //  mean_phi ~ uniform(0, 1);
+  //  mean_p ~ uniform(0, 1);
+  //  gamma ~ uniform(0, 1);
+  //  sigma ~ uniform(0, 5);
+  // In case of using a weakly informative prior
```
bob-carpenter: same comment as before on this extra comment---it's not necessary in the code as it's not documenting the code itself

BPA/Ch.10/js_tempran.stan
```diff
@@ -80,103 +149,52 @@ transformed data {
...
-        lp[t] <- bernoulli_log(1, prod(qgamma[1:(t - 1)]))
-          + bernoulli_log(1, gamma[t])
-          + bernoulli_log(0, p[i, t])
-          + bernoulli_log(1, chi[i, t]);
-      // Never entered
-      lp[n_occasions + 1] <- bernoulli_log(1, prod(qgamma));
-      increment_log_prob(log_sum_exp(lp));
-    }
-  }
+  jolly_seber_lp(M, n_occasions, y, first, last,
```
bob-carpenter: If you named the function `jolly_seber_lpmf` you could write

```
M ~ jolly_seber(n_occassions, y, first, last, p, phi, gamma, chi);
```

No efficiency difference and which you find more natural is of course a matter of taste.  For the `jolly_seber_lpmf` version, you'd accumulate the densities as you go.  So it probably won't be quite as efficient as the way you have it because we don't yet have a way to drop constants in anything other than built-in (not user defined) sampling statements.

BPA/Ch.10/js_ms.stan
```diff
@@ -16,133 +16,129 @@ data {
...
-      cprob[t] <- gamma[t] * prod(qgamma[1:(t - 1)]);
-    } //t
-    psi <- sum(cprob);
-    for (t in 1:(n_occasions - 1)) {
-      b[t] <- cprob[t] / psi;
-    } //t
+    qgamma = 1.0 - gamma;
+    cprob[1] = gamma[1];
+    for (t in 2:n_occ_minus_1)
+      cprob[t] = gamma[t] * prod(qgamma[1:(t - 1)]);
```
ito4303: Thank you for your suggestion. I incorporated the function into the models.

BPA/Ch.10/js_super.stan
ito4303: I found that u[i] was not needed for the calculation, and removed it.

BPA/Ch.10/js_tempran.stan
ito4303: I removed the unnecessary comments.

BPA/Ch.10/js_tempran.stan
ito4303: I left the function unchanged because the change might become a bit complicated for me.

ito4303:
I updated the models. If they look good, please merge them.

PR #92, opened 2017-02-01
BPA/Ch.12 update

ito4303:
I updated the models in BPA/Ch.12. I adopted bivariate Poisson implementation into the N-miture models in Section 12.3, and it made the calculation very fast. Would you review the code?

### DIFF
--- a/BPA/Ch.12/Nmix2.stan
+++ b/BPA/Ch.12/Nmix2.stan
@@ -18,27 +18,24 @@ functions {
    *
    * return Log probability
   */
-  real bivariate_poisson_lpmf(int[] n, real log_lambda, real[] logit_p) {
+  real bivariate_poisson_log_lpmf(int[] n, real log_lambda, real[] p) {
     real s[min(n) + 1];
-    real lambda = exp(log_lambda);
-    real p[2] = inv_logit(logit_p);
 
-    if (lambda < 0) {
-      reject("lambda must be non-negative.");
-    } else if (p[1] < 0 || p[1] > 1 || p[2] < 0 || p[2] > 1) {
+    if (size(n) != 2)
+      reject("Size of n must be 2.");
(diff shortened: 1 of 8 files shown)


BPA/Ch.12/Nmix2.stan
```diff
@@ -1,145 +1,171 @@
...
-        pr <- softmax(lp[i, k]);
-        N[i, k] <- categorical_rng(pr) - 1;
-        for (j in 1:T) {
-          p[i, j, k] <- inv_logit(logit_p[i, j, k]);
+    // Initialize N and ik_p
+    N = rep_array(0, R, 7);
+    ik_p = rep_matrix(0, R, 7);
+    for (i in 1:R) {
+      // Initialize E and E_new
+      E[i] = rep_matrix(0, T, 7);
```
bob-carpenter: If there are efficiency issues, this will be more efficient if you set

```
E[i] = E[i - 1];
```

for cases where `i > 1`

bob-carpenter:
If you're OK with this as is, I'm happy to merge it whenever.  Thanks for all the updates.

BPA/Ch.12/Nmix2.stan
```diff
@@ -1,145 +1,171 @@
...
+      num_obs_site[k] = num_obs_site[k] + (y[i, 1, k] != -1);
   }
 }
 
 parameters {
-  vector[7] alpha_lam;
+  vector<upper=7>[7] alpha_lam; // Constraint for stability
   vector[7] beta;
-  vector[R] eps_raw;
+  vector<upper=7>[R] eps_raw;   // Constraint for stability
```
ito4303: I think the instability is mainly due to the insufficient information of the data used in the book (this is discussed in the book). With this data, lambda can be very huge when both of p[1] and p[2] become very small during the calculation of bivariate Poisson.

ito4303:
Thank you for your review. I updated the models. If they look good, please merge them.

PR #94, opened 2017-02-07
BPA/Ch.13 update

ito4303:
I update the models in BPA/Ch.13. This is the last chapter. Would you review the code?

### DIFF
--- a/BPA/Ch.13/owls_ms1.stan
+++ b/BPA/Ch.13/owls_ms1.stan
@@ -18,27 +18,27 @@ transformed parameters {
   simplex[3] phi[R];            // State vector
   simplex[3] p[3, T];           // Observation matrix
 
-  p3 <- beta / sum(beta);       // Induce Dirichlet prior
+  p3 = beta / sum(beta);       // Induce Dirichlet prior
 
   // Define state vector
   for (s in 1:R) {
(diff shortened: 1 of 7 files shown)


BPA/Ch.13/owls_ms1.stan
```diff
@@ -18,27 +18,27 @@ transformed parameters {
...
-    p[1, t, 1] <- 1.0;
-    p[1, t, 2] <- 0.0;
-    p[1, t, 3] <- 0.0;
-    p[2, t, 1] <- 1.0 - p2;
-    p[2, t, 2] <- p2;
-    p[2, t, 3] <- 0.0;
-    p[3, t, 1] <- p3[1];
-    p[3, t, 2] <- p3[2];
-    p[3, t, 3] <- p3[3];
+    p[1, t, 1] = 1;
```
bob-carpenter: Your way may be clearer, but you could also do this via:

```
p[1, 1:T, 1] <- rep_array(1, T);
p[1, 1:T, 2] <- rep_array(0, T);
```
and so on.  I'm not 100% sure this'll work, but it should.

It's more efficient this way as you do less arithmetic copying than continually recalculating `1 - p2` and creating distinct zero variables that aren't tied back to any parameters (so can all be shared).

ito4303:
Thank you for your review. I updated the models using rep_array. Please merge them if they look good.

PR #96, opened 2017-02-14
BPA/Ch.13 Update a site-occupancy model

ito4303:
I updated a site-occupancy model to improve accuracy of an estimate. Would you review the code?

### DIFF
--- a/BPA/Ch.13/site-occ_cov.stan
+++ b/BPA/Ch.13/site-occ_cov.stan
@@ -55,17 +55,19 @@ model {
 }
 
 generated quantities {
-  int<lower=0> occ_fs;    // Number of occupied sites
-  int z[R];
-
-  // This code fully simulate the states without condition of
-  // observed y, so that the result will disperse wider than
-  // that of the book.
+  int occ_fs = occ_obs; // Number of occupied sites
+  // Number of sites observed as occupied
+  // + Number of sites occupied and not observed
   for (i in 1:R) {
-    real psi;
+    if (sum_y[i] == 0) {
+      int z;
+      real psi = inv_logit(logit_psi[i]);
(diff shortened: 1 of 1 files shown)


BPA/Ch.13/site-occ_cov.stan
```diff
@@ -55,17 +55,19 @@ model {
...
   for (i in 1:R) {
-    real psi;
+    if (sum_y[i] == 0) {
+      int z;
+      real psi = inv_logit(logit_psi[i]);
+      vector[T] p = inv_logit(logit_p[i])';
 
-    psi = inv_logit(logit_psi[i]);
-    z[i] = bernoulli_rng(psi);
+      z = bernoulli_rng(psi);
```
bob-carpenter: You can just move the declaration down here and use

```
int z = bernoulli_rng(psi);
```

There's no overhead for local declarations if they're not containers (arrays, vectors, matrices).

ito4303: Unfortunately, the local declaration with `bernoulli_rng()` seems to be unavailable.

It generates the following error.

```
Syntax ERROR, MESSAGE(S) FROM PARSER:

ERROR: random number generators only allowed in transformed data block, generated quantities block or user-defined functions with names ending in _rng; found function=bernoulli_rng in block=local

ERROR at line 65

 63:          real psi = inv_logit(logit_psi[i]);
 64:          vector[T] p = inv_logit(logit_p[i])';
 65:          int z = bernoulli_rng(psi);
                                        ^
 66:    
```

bob-carpenter: Sorry about that---that's a bug in the parser where it's being too strict on scope.  It's fixed on the develop branch for 2.15.  But I'll just merge this for now---nothing wrong with the way you have it.

PR #100, opened 2017-03-19
Fixes for issue 99, BPA examples

mikemeredith:
Changes to generated quantities in Stan models to use conditional (on
non-detection) values for psi or omega as discussed in Issue 99. Also
typo in Ch.07/chs_mnl.stan

### DIFF
--- a/BPA/Ch.06/M0.stan
+++ b/BPA/Ch.06/M0.stan
@@ -46,7 +46,9 @@ model {
 }
 
 generated quantities {
-  int<lower=C> N;
+  int<lower=C, upper=M> N;
+  real omega_nd;  // prob present given never detected
+  omega_nd = (omega * (1 - p)^T) / (omega * (1 - p)^T + (1 - omega));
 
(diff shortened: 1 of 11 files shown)


mikemeredith:
For the Ch.06 models, I checked against the JAGS output, and in fact both the old and new results are close to JAGS.
I wrote a script to check that all the models in the BDA directory would compile, and that threw up an error with` Ch.07/cjs_mnl.stan`, "n_occ_minus_1" not found, which seems clearly a typo for "n_occasions_minus_1"; reject that if not the case.

ito4303:
> that threw up an error withCh.07/cjs_mnl.stan, "n_occ_minus_1" not found, which seems clearly a typo for "n_occasions_minus_1"; reject that if not the case.

This model was accidentally dropped from my checking script, and also should be fixed.

bob-carpenter:
@ito4303  Should this be merged?  I just realized it's just been sitting here.

ito4303:
Yes. Please.

PR #101, opened 2017-03-21
Changes_for_dorazio-et-al-knitr

mikemeredith:
@bob-carpenter  The estimation of `E_N_2` is ok, it uses the proper conditional probability, unlike some of the models in the BPA translation.

The big issue is the equation `z[i,j] = I(x[i,j] > 0)` and the accompanying statement that "The values of `z[i,j]` for `i < n` are observed, ..." (lines 138, 142). This contradicts the quote from Dorazio et al on line 160. The equations from line 164 onwards and the Stan code are correct, but this needs fixing as (a) it misinforms the reader about the Dorazio et al model (or suggests that you don't understand it!), and (b) it is *not* the definition of z[i,j] used subsequently.

The paper uses the term "abundance", while it's actually occupancy or incidence that is being modelled. There are multi-species abundance models, eg, Yamaura et al (2011) J Applied Ecol. 48, 67-75, but this is not one of them. Suggest changing "abundance" to "occupancy" or "occurence" throughout to avoid confusion.

Size of supercommunity, `S`: From the Supplementary Materials, Dorazio et al used `S = 128`, adding 100 undetected species to the 28 detected; 100 is needed for the bird example in their paper, but is over-kill for the butterflies, 50 total is fine. Suggest adding a comment after the replot of their Fig. 6 (line 458) on the adequacy of 50, as with your own data you would not be able to check with Dorazio et al.

The modified .Rmd file knits ok, with warnings about a bunch of deprecated stuff in the Stan model description. I also got warnings about unused functions and typedefs, and low estimated Bayesian Fraction of Missing Information.

Thanks very much for doing this case study!

### DIFF
--- a/knitr/dorazio-royle-occupancy/dorazio-et-al-knitr.Rmd
+++ b/knitr/dorazio-royle-occupancy/dorazio-et-al-knitr.Rmd
@@ -49,7 +49,7 @@ for (i in 1:n) {
   }
 }
 K <- 18;   # read from paper, figure 5
-S <- 50;   # guessing from plot in figure 6
+S <- 50;   # more than adequate judging from plot in figure 6
 head(x_csv);
 ```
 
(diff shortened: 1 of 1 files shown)


bob-carpenter:
@mikemeredith I posted the new version online and acknowledged you for helping out.  Thanks again!

PR #103, opened 2017-04-12
BPA: Fix mistakes in R files

ito4303:
Correspondences between R and Stan files were wrong in Chapter 7.


bob-carpenter:
I merged this, then invited you to join the team for example-models.  Then you should be able to merge your own requests.  I'm still happy to review, but you should also be able to merge directly.

PR #107, opened 2017-06-26
Update models in Ch.06 of BPA

ito4303:
Non-centered parameterization is applied for hierarchical models.

### DIFF
--- a/BPA/Ch.06/6.2.4_Model_Mh.R
+++ b/BPA/Ch.06/6.2.4_Model_Mh.R
@@ -12,6 +12,7 @@ set.seed(1234)
 ## The data generation code is in bpa-code.txt, available at
 ## http://www.vogelwarte.ch/de/projekte/publikationen/bpa/complete-code-and-data-files-of-the-book.html
 stan_data <- read_rdump("Mh.data.R")
+
 ## Initial values
 inits <- function() list(mean_p = 0.5, sigma = runif(1, 0.5, 0.9))
 
@@ -29,7 +30,6 @@ out <- stan("Mh.stan",
             data = stan_data, init = inits, pars = params,
             chains = nc, iter = ni, warmup = nb, thin = nt,
             seed = 1,
-            control = list(adapt_delta = 0.8),
             open_progress = FALSE)
 
 ## Summarize posteriors
--- a/BPA/Ch.06/6.2.4_Model_Mh.R
+++ b/BPA/Ch.06/6.2.4_Model_Mh.R
@@ -12,7 +12,6 @@ set.seed(1234)
 ## The data generation code is in bpa-code.txt, available at
 ## http://www.vogelwarte.ch/de/projekte/publikationen/bpa/complete-code-and-data-files-of-the-book.html
 stan_data <- read_rdump("Mh.data.R")
-
 ## Initial values
 inits <- function() list(mean_p = 0.5, sigma = runif(1, 0.5, 0.9))
 
@@ -20,9 +19,9 @@ inits <- function() list(mean_p = 0.5, sigma = runif(1, 0.5, 0.9))
 params <- c("N", "mean_p", "sigma", "omega")
 
 ## MCMC settings
-ni <- 15000
-nt <- 10
-nb <- 5000
+ni <- 10000
+nt <- 9
+nb <- 1000
 nc <- 4
 
 ## Call Stan from R
@@ -30,9 +29,8 @@ out <- stan("Mh.stan",
             data = stan_data, init = inits, pars = params,
             chains = nc, iter = ni, warmup = nb, thin = nt,
             seed = 1,
-            control = list(adapt_delta = 0.95),
+            control = list(adapt_delta = 0.8),
             open_progress = FALSE)
-## Note: There may be divergent transitions after warmup.
 
 ## Summarize posteriors
 print(out, digits = 3)
--- a/BPA/Ch.06/Mt.stan
+++ b/BPA/Ch.06/Mt.stan
@@ -27,30 +27,21 @@ model {
   //  p ~ uniform(0, 1);
 
   // Likelihood
-  for (i in 1:M) {
-    real lp[2];
-
-    if (s[i] > 0) {
+  for (i in 1:M)
+    if (s[i] > 0)
       // z[i] == 1
       target += bernoulli_lpmf(1 | omega)
               + bernoulli_lpmf(y[i] | p);
-    } else { // s[i] == 0
-      // z[i] == 1
-      lp[1] = bernoulli_lpmf(1 | omega)
-            + bernoulli_lpmf(0 | p);
-      // z[i] == 0
-      lp[2] = bernoulli_lpmf(0 | omega);
-      target += log_sum_exp(lp[1], lp[2]);
-    }
-  }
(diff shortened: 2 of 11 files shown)


bob-carpenter:
I didn't understand the non-centering in `BPA/Ch.06/Mh.stan`.  You have (equivalently---I translated to use compound declare/define):

```
parameters {
  real<lower=0,upper=1> omega;      // Inclusion probability
  real<lower=0,upper=1> mean_p;     // Mean detection probability
  real<lower=0,upper=5> sigma;
  vector[M] sigma_raw;
  ...
transformed parameters {
  vector[M] eps = logit(mean_p) + sigma * sigma_raw;
  ...
model {
  ...
  sigma_raw ~ normal(0, 1);
```

I think it's partly confusing naming and partly an unusual approach to centering.  I think what you really want to do if you want to parameterize in terms of log odds is to use

```
parameters {
  real mu_eps;
  vector[M] eps_raw;
  real<lower=0> sigma_eps;
  ...
transformed parameters {
  vector[M] eps = mu_eps + eps_raw * sigma_eps;
  ...
model {
  eps_raw ~ normal(0, 1);
```

You could alternatively draw `eps_raw` from a `logistic(0, 1)`, which gives you the same distribution as a `uniform(0, 1)` variate that is log-odds transformed.

If you're going to change the models from the way they were coded in the book, I'd suggest not using those interval constrained parameters as they will mess up both computation and estimation if there would otherwise be any probability mass near the boundaries.

bob-carpenter:
I'd also suggest rewriting the likelihood (which doesn't include `sigma_raw`) as:

```
for (i in 1:M) {
  if (y[i] > 0)
    target += bernoulli_lpmf(1 | omega)
                + binomial_logit_lpmf(y[i] | T, eps[i]);
  else 
    target += log_sum_exp(bernoulli_lpmf(1 | omega)
                            + binomial_logit_lpmf(0 | T, eps[i]),
                          bernoulli_lpmf(0 | omega));
}
```

BPA/Ch.06/6.2.4_Model_Mh.R
```diff
@@ -12,27 +12,25 @@ set.seed(1234)
...
+nb <- 1000
 nc <- 4
 
 ## Call Stan from R
 out <- stan("Mh.stan",
             data = stan_data, init = inits, pars = params,
             chains = nc, iter = ni, warmup = nb, thin = nt,
             seed = 1,
-            control = list(adapt_delta = 0.95),
+            control = list(adapt_delta = 0.8),
```
bob-carpenter: `adapt_delta = 0.8` is the default, so you can just skip it here and in future uses.

BPA/Ch.06/Mt.stan
```diff
@@ -49,7 +49,7 @@ generated quantities {
   int<lower=C> N;
   real pr;        // prob never captured given present
   real omega_nd;  // prob present given never captured; same for all animals
-  
+
```
bob-carpenter: You can use compound declare/define for all these, which really makes the programs a lot simpler.

BPA/Ch.06/Mt.stan
```diff
@@ -49,7 +49,7 @@ generated quantities {
...
   pr = prod(rep_vector(1.0, T) - p);
```
bob-carpenter: This can just be `prod(1 - p)` if `p` is a vector---scalars in vector addition and subtraction get broadcast and the result is more efficient and also easier to read.

BPA/Ch.06/Mtbh.stan
```diff
@@ -23,27 +23,25 @@ parameters {
...
+  eps = sigma * sigma_raw;
   for (j in 1:T)
     alpha[j] = logit(mean_p[j]); // Define logit
   for (i in 1:M) {
-    real logit_p;
 
     // First occasion: no term for recapture (gamma)
-    logit_p = alpha[1] + eps[i];
-    p[i][1] = inv_logit(logit_p);
+    logit_p[i, 1] = alpha[1] + eps[i];
```
bob-carpenter: You can do all this with vectorization:

```
logit_p[ , 1] = alpha[1] + eps;
```

BPA/Ch.06/Mtbh.stan
```diff
@@ -23,27 +23,25 @@ parameters {
...
 
     // All subsequent occasions: includes recapture term (gamma)
-    for (j in 2:T) {
-      logit_p = alpha[j] + eps[i] + gamma * y[i, j - 1];
-      p[i][j] = inv_logit(logit_p);
-    }
+    for (j in 2:T)
+      logit_p[i, j] = alpha[j] + eps[i] + gamma * y[i, j - 1];
```
bob-carpenter: Same here:

```
logit_p[ , j] = alpha[j] + epsi + gamma * y[ , j - 1];
```

Then you shouldn't need to loop over `i`.

BPA/Ch.06/Mtbh.stan
```diff
@@ -80,10 +77,14 @@ model {
 }
 
 generated quantities {
+  vector<lower=0,upper=1>[T] p[M];
   int<lower=0,upper=1> z[M];
   int<lower=C> N;
 
   for (i in 1:M) {
+    for (j in 1:T)
```
bob-carpenter: It's more efficient to vectorize this whole loop the other way:

```
for (j in 1:T)
  p[ , j] = inv_logit(logit_p[ , j]);
```

BPA/Ch.06/Mth.stan
```diff
@@ -22,18 +22,20 @@ parameters {
...
-  vector<lower=0,upper=1>[T] p[M];
+  vector[T] logit_p[M];
 
+  eps = sigma * sigma_raw;
   for (j in 1:T)
     mean_lp[j] = logit(mean_p[j]); // Define logit
   for (i in 1:M)
     for (j in 1:T)
-      p[i][j] = inv_logit(mean_lp[j] + eps[i]);
+      logit_p[i, j] = mean_lp[j] + eps[i];
```
bob-carpenter: This can also be vectorized.

BPA/Ch.06/Mth.stan
```diff
@@ -67,16 +67,22 @@ model {
 }
 
 generated quantities {
+  vector<lower=0,upper=1>[T] p[M];
   int<lower=0,upper=1> z[M];
   int<lower=C> N;
 
   for (i in 1:M) {
+    for (j in 1:T)
```
bob-carpenter: Same comments in these loops as in the previous ones.  And if possible, pulling this out into a shared function as you've done elsewhere might make the structure across models a little easier to see.

BPA/Ch.06/Mtbh.stan
ito4303: This was not applicable because y[ , ] was an integer array.

ito4303:
Thank you for your comments. I fixed the mistakes and updated models using vectorization and compound declaration.

BPA/Ch.06/Mtbh.stan
bob-carpenter: If you only use `y` as a predictor, as in the definition of `logit_p`, then you can declare it as a `matrix`.  You can assign integers to matrices in data.

PR #112, opened 2017-07-10
Car iar poisson edits dan

dpsimpson:
Hi,

I've added some a "best practice" parameterisation to the example and a few more references.  Everything runs (with tonnes of C++ warnings).  I didn't suppress the warnings (even though they're basically meaningless) because Mitzi didn't.

Let me know if anything doesn't make sense.  Or anything should be changed.

Best wishes

Dan


bob-carpenter:
Sorry about the C++ errors.  The C++ warnings are coming Boost;  we would have to hack the Boost lib imported from Boost to suppress these erroneous warnings.  But we're reluctant to modify Boost just to do this.

mitzimorris:
this is great!  merging!

dpsimpson:
Brilliant!

I think adding
  message=F, warning=F
To the code chunks will suppress the messages, but I'll leave that up to
Mitzi.

D

betanalpha:
What you want to do is modify your makevar file before running knitr.
See http://mc-stan.org/misc/warnings.html#compiler-warnings <http://mc-stan.org/misc/warnings.html#compiler-warnings>.

There are a few formatting issues here and there (mostly to do with
line breaks) but I figured they’ll be cleaned up in future iterations.

mitzimorris:
cleaning up now.
also made a version of your bym2 model for the Brooklyn data which has no offset.
changing the comments in the model on the 'sigma' and 'rho' params to match
what these guys are doing.

once the Rmd knits, will run the inla functions on the Brooklyn data.

there's another directory which has the R code to run the Cook, Gelman, and Rubin
procedure to validate the model by replication, and I can run it over this model as well -
for a given spatial structure, the scaling factor is constant, correct?

cheers,
Mitzi



> What you want to do is modify your makevar file before running knitr.
> See http://mc-stan.org/misc/warnings.html#compiler-warnings
> <http://mc-stan.org/misc/warnings.html#compiler-warnings>.
>
> There are a few formatting issues here and there (mostly to do with
> line breaks) but I figured they’ll be cleaned up in future iterations.
>

dpsimpson:
That's correct. The scaling is a function of the graph.

Let me know if you need any help making the priors match in inla (or with
anything else)

D

mitzimorris:
rereading Riebler et al which is beginning to make sense.
I would like to add your name as author of this case study - is that OK?

how is Toronto?

we'll be in Paris from July 23 through Sept 13th.
when we get back we've got a visit to Halifax promised sometime
during late September/early October.  after that, I'd like to get going
on my plan to move to higher ground, both politically and geographically.
that iceberg thing has me freaked.

cheers,
Mitzi


> That's correct. The scaling is a function of the graph.
>
> Let me know if you need any help making the priors match in inla (or with
> anything else)
>
> D
>

dpsimpson:
Hi Mitzi,
You can add me as an author if you want, but it feels like I'm just swanning in to your hard work and taking credit (after all, you did 95% of it!).  Maybe listing me as a "contributor" would be more fair?

I thiknk I'm in NY in late September so hopefully I'll see you there (and you're always welcome here for a visit.  Also there are jobs in Toronto...)

D

PR #113, opened 2017-07-15
Dan inla stan

dpsimpson:
I have a comparision with INLA that has matching priors etc.  you get basically the same answer.

While doing this, I found an error in how I used the scaling_factor in the .stan file I wrote (it needed to scale the standard deviation properly!).  This is fixed now.

Some of the differences in the results disappear with a longer Stan chain and with the slower (ie non-default) INLA commands that have been commented out in the file.


bob-carpenter:
Thanks, @dpsimpson, this is fantastic.  It should go a long way toward assuaging the BUGS users fears about Stan.

dpsimpson:
Procrastination is sometimes a wonderful thing @bob-carpenter!

(Today's activity involves actively disliking Eigen's documentation)

bob-carpenter:
That's Andrew's strategy---productive procrastination.  When you next want to procrastinate, our matrix library could use some tuning at the C++ level :-)

I really like Eigen's overview docs and how-to guides for the basic matrix functionality.  Their developer-facing stuff like the how-to-write functions and the doxygen doc usually leave me confused.  The code itself is very hard to read unless you're way into templated C++.

dpsimpson:
Today it was "I know there's got to be a reverse iterator but I had to grep through the rep to find it"...

More annoyingly, I can't work out the return type of a function...

PR #126, opened 2018-02-21
revised to only use two mdivide_ calls

Sycor4x:
this should be more efficient since mdivide_ is expensive.

### DIFF
--- a/misc/gaussian-process/gp-predict-analytic.stan
+++ b/misc/gaussian-process/gp-predict-analytic.stan
@@ -14,9 +14,9 @@ functions{
   
       {
         matrix[N1, N1] L_K;
-        vector[N1] K_div_y1;
+        vector[N1] L_div_y1;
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks.  I hope @rtrangucci or @betanalpha or @avehtari can review.  This one's making my head hurt from the linear algebra.

Sycor4x:
Over the weekend I'll take a look at the other GP scripts and see if similar improvements may be made.

avehtari:
A good catch. Yes, it's correct. For big n, Cholesky will still dominate, but there shouldn't be any downsides for this change. Do you want to add other changes to the same pull request or make a new one for those?

Sycor4x:
Your point about the Cholesky factorization dominating the computational cost is well-taken.

Please merge this one. Once I make review of the other GP scripts, I'll make a new PR to group together any of my possible future efficiency improvements.

PR #128, opened 2018-04-06
Fix some notation and formatting.

mkcor:
Hi @mitzimorris!

Hope this finds you well.

I'm not entirely sure of my changes, which is why I checked "Allow edits from maintainers."

I understand that there are $n$ regions and we identify a region by $i$ (and, say, another region by $j$). So $i$ runs from $1, ..., n$.

But, apparently, there are also "$n$ different areal units of a region" (in a region? in each region?); then I would understand $n_i$ to be an areal unit of region $i$.

Does $n_i = 1, ..., n$? It would seem odd that this would be the same $n$ (number of regions vs. number of areal units per region).

Am I completely off? Please let me know!

I've also noticed that, starting at section "Adding an ICAR component to a Stan model," $n$ (which one?) seems to become $N$ while, earlier on, $N$ is used to denote the normal distribution.

Thank you,
Marianne

### DIFF
--- a/knitr/car-iar-poisson/icar_stan.Rmd
+++ b/knitr/car-iar-poisson/icar_stan.Rmd
@@ -52,7 +52,7 @@ $\mathbf{\phi} = ({\phi}_1, \ldots, {\phi}_n)^T$.
 
 For CAR models, spatial relationship between the $n$ areal units
 are represented as an adjacency matrix $W$ with dimensions $n \times n$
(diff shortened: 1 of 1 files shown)


mkcor:
Hi @mitzimorris,

Thanks for sharing!

One of @bayesways's and my projects (when we find the time...) is to translate this case study into Python (https://github.com/bayesways/spatial_stan), so that we can learn along the way and (maybe? hopefully?) contribute something useful to the community.

Cheers,
Marianne

PR #129, opened 2018-04-12
Bugfix/127 compile models 2.17.1+

mitzimorris:


mitzimorris:
fixed what could be fixed; all models compile except for 3 models in ARM/Ch.21
requesting signoff from @seantalts

PR #134, opened 2018-05-09
Fix typo.

alessandro-gentilini:


bob-carpenter:
Thanks.  What exactly was being fixed here?  I couldn't tell from a glance through the diff.

Even for small changes, we ask for the name of the copyright holder for the contribution.

alessandro-gentilini:
I changed the word dyanmics to dynamics.

PR #143, opened 2018-11-11
Simplify LDA input parameterization

gokceneraslan:
I tried to simplify LDA input representation by using a simple `M x V` matrix of word frequencies where M and V represent number of documents and words. In the model, now instead of iterating over all words of all documents, iterations are over each element of the `M x V` matrix.


bob-carpenter:
Thanks for submitting.  I've been out for a while, so haven't been able to review this, but I'll get to it ASAP.

misc/cluster/lda/corr-lda.stan
```diff
@@ -38,10 +36,16 @@ model {
...
-      gamma[k] <- log(theta[doc[n],k]) + log(phi[k,w[n]]);
-    increment_log_prob(log_sum_exp(gamma));  // likelihood
+  for (i in 1:M) {
+    for (j in 1:V) {
+      int count = corpus[i,j];
+      real gamma[K];
+      if (count > 0) {
+        for (k in 1:K) {
+          gamma[k] = (log(theta[i,k]) + log(phi[k,j]))*count;
+        }
```
bob-carpenter: Rather than looping to define `gamma`, a one-liner will do it:

```
gamma = count * (log(theta[i, ]) + log(phi[, j]);
```

The loop works for the `count == 0` case (though a bit ineffciently) but presumably there aren't any zero-length documents in well-formed data sets, so I'd just write replace this whole loop with:

```
for (i in 1:M)
  for (j in 1:V)
    target += log_sum_exp(count * (log(theta[i, ]) + log(phi[, j]));
```

`increment_log_prob` is deprecated---this model has been around for a while without being updated.

misc/cluster/lda/corr-lda.stan
```diff
@@ -2,28 +2,26 @@ data {
...
   simplex[V] phi[K];         // word dist for topic k
 }
 transformed parameters {
   simplex[K] theta[M];       // simplex topic dist for doc m
   cov_matrix[K] Sigma;       // covariance matrix
   for (m in 1:M)
-    theta[m] <- softmax(eta[m]);
+    theta[m] = softmax(eta[m]);
   for (m in 1:K) {
-    Sigma[m,m] <- sigma[m] * sigma[m] * Omega[m,m];
```
bob-carpenter: Thanks for replacing `<-`  --- that would be a good change for the original model, as well as replacing `increment_log_prob` with `target +=`.

`square(sigma[m])` will be a bit more efficient, as would be `sigma[m]^2`.

bob-carpenter:
Oh, and I'd suggest adding suffixes to existing model names like `_counts` to indicate you're taking sufficient stats rather than the raw data.

gokceneraslan:
> Oh, and I'd suggest adding suffixes to existing model names like `_counts` to indicate you're taking sufficient stats rather than the raw data.

You mean adding `_counts` to the new model? Because it's the one uses counts.

bob-carpenter:
Anything to distinguish the way in which data is coded in the two approaches.  So yes, I meant keeping <foo>.stan as is and adding <foo>_counts.stan or something similar for the sufficient stats version.

PR #149, opened 2019-05-15
updated mark recapture examples

rerickson-usgs:
@bob-carpenter I am trying to use these for a different project. I replaced `<-` with `=` and changed the code to use `target +=` syntax.


rerickson-usgs:
@bob-carpenter Any chance you've had time to look at this proposed merge?

seantalts:
Thanks!

PR #156, opened 2019-10-03
Upload causal case study 1

joonho112:


bob-carpenter:
I'm just going to merge this---we treat the example-models repo very loosely.

PR #158, opened 2019-10-15
HMM fix

imadmali:
There are some minor issues with the HMM example models. One of the models defines a variable called `best_total_logp` that is never used and there are some deprecated features that are used (e.g. assignment operators). The purpose of this pull request is to fix these issues.

Note that the definition of `best_total_logp` also occurs in the documentation but I am not sure which repository that is created in and where that should be fixed.


imadmali:
@bob-carpenter where is the Stan User's Guide documentation generated? (The `best_total_logp` declaration also needs to be fixed there.)

PR #160, opened 2019-11-07
Update  12.2.2_Covariates.R

hneyster:
Update 12.2.2_Covariates.R to include the covariate data (X) in the stan call

### DIFF
--- a/BPA/Ch.12/12.2.2_Covariates.R
+++ b/BPA/Ch.12/12.2.2_Covariates.R
@@ -31,7 +31,7 @@ inits <- lapply(1:nc, function(i)
 
 ## Call Stan from R
 out <- stan("binmix_cov.stan",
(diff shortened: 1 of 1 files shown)


BPA/Ch.12/12.2.2_Covariates.R
```diff
@@ -31,7 +31,7 @@ inits <- lapply(1:nc, function(i)
 
 ## Call Stan from R
 out <- stan("binmix_cov.stan",
-            data = list(y = y, R = R, T = T, K = 100),
+            data = list(y = y, X=X, R = R, T = T, K = 100),
```
ito4303: For consistency of coding style,  `X = X` (spaces around operators) would be preferable.

hneyster: Good point. Fixed now.

PR #164, opened 2020-03-11
Updated the stan syntax to use log_mix

jarnefeltoliver:
Now using log_mix which allowed the deletion of transformed parameters block.

### DIFF
--- a/basic_estimators/normal_mixture.stan
+++ b/basic_estimators/normal_mixture.stan
@@ -9,17 +9,11 @@ parameters {
   real<lower=0,upper=1> theta;
   real mu[2];
 }
-transformed parameters {
-  real log_theta;
-  real log_one_minus_theta;
-
-  log_theta = log(theta);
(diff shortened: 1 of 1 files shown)


jarnefeltoliver:
Okay, thanks I will keep those in mind!

Oliver
________________________________
Lähettäjä: Bob Carpenter <notifications@github.com>
Lähetetty: 11. maaliskuuta 2020 18:09:31
Vastaanottaja: stan-dev/example-models
Kopio: Järnefelt Oliver; Author
Aihe: Re: [stan-dev/example-models] Updated the stan syntax to use log_mix (#164)


@bob-carpenter approved this pull request.

Looks good. Thanks.

I'm going to merge this, but if you want to keep going:

  1.  mu sampling should be vectorized
  2.  theta sampling should be removed
  3.  1 instead of 1.0
  4.  spaces after commas and around operators (other than :)

PR #176, opened 2020-05-11
Updated reduce_sum tutorial for cmdstanr threading interface chanes (Issue #175)

### LINKED ISSUE #175: reduce_sum case study will be out of date once new cmdstanr pull gets merged
The pull in question is here (with information on the problem and the fix): https://github.com/stan-dev/cmdstanr/pull/144#issuecomment-626840721

The difference is in how threading is used.

bbbales2:
Fixes #175 

The associated website pull request is [here](https://github.com/stan-dev/stan-dev.github.io/pull/115).

Edit: Added link to website repo pull

### DIFF
--- a/knitr/reduce-sum/reduce_sum_tutorial.Rmd
+++ b/knitr/reduce-sum/reduce_sum_tutorial.Rmd
@@ -1,6 +1,6 @@
 ---
 title: "Reduce Sum: A Minimal Example"
-date: "17 April 2020"
+date: "11 May 2020"
 output: html_document
 ---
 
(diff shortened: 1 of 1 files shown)


bbbales2:
> Do you want to also mentioned that CmdStanR is still unreleased, so people don't think we're breaking released interfaces?

I don't think it practically matters if we called it unreleased. It's released in the sense that we've told people to use it.

Go ahead and merge these just so we don't forget?

jgabry:
Yeah that's true. I'll merge now

PR #178, opened 2020-05-13
Changed num_cores/num_chains to cores/chains

bbbales2:
Switched the tutorial to use cores/chains instead of num_cores/num_chains.

### DIFF
--- a/knitr/reduce-sum/reduce_sum_tutorial.Rmd
+++ b/knitr/reduce-sum/reduce_sum_tutorial.Rmd
@@ -1,6 +1,6 @@
 ---
 title: "Reduce Sum: A Minimal Example"
-date: "11 May 2020"
+date: "13 May 2020"
 output: html_document
 ---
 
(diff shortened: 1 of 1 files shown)


knitr/reduce-sum/reduce_sum_tutorial.Rmd
```diff
@@ -286,12 +287,13 @@ Run and time the model with:
...
 time1 = system.time(fit1 <- logistic1$sample(redcard_data,
-                                             num_cores = 4,
-                                             num_chains = 4,
+                                             cores = 4,
+                                             chains = 4,
                                              refresh = 1000))
 
 time1
 ```
+**Note:** Older versions of cmdstanr use `num_cores` and `num_chains` instead of `cores` and `chains`. If you get an error, [update](https://mc-stan.org/cmdstanr/) `cmdstanr`.
```
jgabry, line 296: There should just be a deprecation warning and not an error, but I guess eventually we will get rid of the deprecation warning and make it an error so fine by me to keep this Note.

Issue #19, opened 2015-02-16
Stacks example


avehtari:
On page
https://github.com/stan-dev/example-models/wiki/BUGS-Examples-Sorted-Alphabetically
under "Stacks: "robust regression"" the link to "Stan code" is
https://github.com/stan-dev/example-models/blob/master/bugs_examples/vol1/stacks/stacks.stan
which gives 404. 
Unlike other examples, Stacks has 6 different Stan code files (with different residual models) and none of them is named stacks.stan. To fix this there should be 6 links to these different models, or just a link to the corresponding "Stan files" folder.

PR #185, opened 2020-07-27
Added model code and data to convert-odes case study (design-doc #19)

bbbales2:
And updated the links in the case study to point at this repo.

The links won't work until this goes in though.


knitr/convert-odes/one_comp_mm_elim_abs.stan
```diff
@@ -0,0 +1,65 @@
...
+
+model {
+  // Priors
+  k_a ~ cauchy(0, 1);
+  K_m ~ cauchy(0, 1);
+  V_m ~ cauchy(0, 1);
+  sigma ~ cauchy(0, 1);
+
+  // Likelihood
+  C ~ lognormal(log(mu_C[, 1]), sigma);
```
wds15, line 58: I need to review the rest... but as this model now lives here... can we change the log-lik to

`  C ~ lognormal(log(mu_C[, 1] + 1E-6), sigma);`

?
This is very useful in most real-world applications and many users bump into this problem which is avoided by shifting by something greater than (but still basically 0) the 1E-8 absolute tolerance as set, since otherwise the ode integrator may return slightly negative numbers which you sort out with this - admittedly - hack. It's a bit like a nugget term for GPs.

bbbales2: It's not necessary for this model, so I don't want to put it in. This tutorial is just about converting from the old interface to the new interface. I don't wanna bring up other stuff.

We need our own ODE methods writeup where this could live.

wds15: ok

PR #187, opened 2020-09-17
Added HMM example

bbbales2:
I saw an HMM question on the forums and thought we could use a copy-paste getting started example for the new interface.


knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,269 @@
...
+}
+```
+
+The `hmm_marginal` function takes the transition matrix and initial state
+directly. In this case the transition matrix needs constructed from `t1`,
+`t2`, and `t3` but that is relatively easy to build.
+
+The measurement model, in contrast, is not passed directly to the HMM function.
+
+Instead, a $KxN$ matrix `log_omega` of log likelihoods is passed in. The
```
charlesm93, line 146: Use $K \times N$. Makes the multiply sign look better.

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,269 @@
...
+```{r echo = TRUE, results = FALSE, message = FALSE}
+model = cmdstan_model("hmm-example.stan")
+fit = model$sample(data = list(N = N, y = y), parallel_chains = 4)
+```
+
+The estimated group means match the known ones:
+```{r}
+fit$summary("mu")
+```
+The estimated of the initial condition is not much more informative than
```
charlesm93, line 192: Either the "estimates of the initial condition" or the "estimated initial condition".

charlesm93 (approved):
This looks good. I'll let you correct the minor points I picked up on. One thing to be mindful about is that the hidden states are not independent. So the marginal probability for each state you get with hmm_hidden_state_prob only conveys part of the information.

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,269 @@
...
+
+A more complete mathematical definition of the HMM model and function interface
+is given in the [Hidden Markov Models](https://mc-stan.org/docs/2_24/functions-reference/hidden-markov-models.html)
+section of the Function Reference Guide.
+
+There are three functions
+
+- `hmm_marginal` - The likelihood of an HMM with the latent discrete states
+integrated out
+- `hmm_latent_rng` - A function to generate random realizations from the posterior of an HMM
```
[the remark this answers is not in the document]
bbbales2: Is it fair to say, "posterior random realizations" or something?

One thing that caught me off guard was I didn't know how to generate things from the prior for this -- like the log_omega argument is there!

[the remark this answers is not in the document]
bbbales2: Also where to say hidden states and where to say latent states? Are they interchangeable or is there a difference somewhere?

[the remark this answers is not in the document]
charlesm93: I guess realizations are not random by definition. Maybe the most precise term would be samples from the posterior. I'm blinking on why we needed the log omega argument there, but it definitely comes up in the algorithm. I would do prior simulations as you did in R.

I'd stick with hidden states to be consistent with the term hidden Markov models. They mean the same thing but latent could be more broadly applied.

jgabry:
> I saw an HMM question on the forums and thought we could use a copy-paste getting started example for the new interface.

Yes, definitely! Thanks for doing this @bbbales2 and @charlesm93 for reviewing. 

I also agree with Charles' suggestion and reason for using "hidden" instead of "latent".

knitr/hmm-example/hmm-example.Rmd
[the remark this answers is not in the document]
bbbales2: Since you're more familiar with the math, do you mind checking the `log_omega` thing?

Now that I think about it I don't understand why it's needed in `hmm_hidden_state_prob` either.

If I want to know the distribution of hidden states at time n of something with transition matrix A and initial state v, I just do v_n^T = v_0^T * A^n.

I'll go through and swap in hidden for latent!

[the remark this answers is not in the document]
bbbales2: Whoops I'm assuming v there is a simplex, so it's not the initial state but the distribution of initial states.

bbbales2:
Did the updates. Check me and I'll make an html and put it up on the website

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+```
+
+## Introduction
+
+CmdStan 2.24 introduced a new interface for fitting Hidden Markov models (HMMs)
+in Stan. This document is intended to provide an example use of this interface.
+
+HMMs model a process where a system probabilistically switches between $K$
+states over a sequence of $N$ points in time. It is assumed that the exact
+state of the system is unknown and must be measured at each state.
```
charlesm93, line 23: "measured" may not be the right term. "infered" makes more sense.

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+
+HMMs are characterized in terms of the transition matrix $\Gamma_{ij}$ (each
+element being the probability of transitioning from state $i$ to state $j$
+between measurements), the types of measurements made on the system (the
+system may emit continuous or discrete measurements), and the initial state
+of the system.
```
charlesm93, line 29: Add "Currently Stan only supports a constant transition matrix. Future versions will allow users to specify a transition matrix for each state."

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+
+Any realization of an HMM is a sequence of $N$ integers in the range $[1, K]$,
```
charlesm93, line 31: "of an HMM" --> "of an HMM's hidden states"

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+however, because of the structure of the HMM, it is not necessary to sample
+the hidden states to do inference on the transition probabilities, the
+parameters of the measurement model, or the estimates of the initial state.
+Estimates of the distribution of states at each measurement time can be
+computed separately.
```
charlesm93, line 36: Maybe it's more precise to write "Posterior samples of the hidden states".

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+}
+```
+
+The trajectory can easily be visualized:
+```{r}
+qplot(1:N, states)
+```
+
+An HMM is useful when the hidden state is not measure directly (if the
+state was measured directly, it wouldn't be hidden).
```
charlesm93, line 90: You just blew my mind away

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,276 @@
...
+  simplex[3] rho;
+  
+  // Parameters of measurement model
+  vector[3] mu;
+  real<lower = 0.0> sigma;
+}
+```
+
+The `hmm_marginal` function takes the transition matrix and initial state
+directly. In this case the transition matrix needs constructed from `t1`,
```
charlesm93, line 146: typo: "needs to be constructed"

charlesm93 (approved):
Please address the comments and then it's fine to merge.

knitr/hmm-example/hmm-example.Rmd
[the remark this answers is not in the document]
bbbales2: Ooooh, I finally realized what I was screwing up here. There is a difference between making draws from the integrated out states of the HMM and generated entirely new HMM trajectories. I added text to indicate that.

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,287 @@
...
+
+A more complete mathematical definition of the HMM model and function interface
+is given in the [Hidden Markov Models](https://mc-stan.org/docs/2_24/functions-reference/hidden-markov-models.html)
+section of the Function Reference Guide.
+
+There are three functions
+
+- `hmm_marginal` - The likelihood of an HMM with the hidden discrete states
+integrated out
+- `hmm_latent_rng` - A function to generate draws of the hidden state that are
```
charlesm93, line 48: "generate _posterior_ draws"

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,287 @@
...
+implicitly integrated out of the model when using `hmm_marginal` (this is
+different than sampling more states with a posterior draw of a transition matrix
+and initial state)
+- `hmm_hidden_state_prob` - A function to compute the distributions of the
```
charlesm93, line 52: "the _posterior_ distributions"

knitr/hmm-example/hmm-example.Rmd
```diff
@@ -0,0 +1,287 @@
...
+
+```{stan, output.var = "", eval = FALSE}
+generated quantities {
+  int[N] y_sim = hmm_latent_rng(log_omega, Gamma, rho)
+}
+```
+
+Note, these are the probabilities and draws of the hidden state that was
+integrated out. This is different than sampling new HMM realizations using
+posterior draws of the initial condition and the transition matrix.
```
charlesm93, line 276: "Note further that the probabilities are the _marginal_ probabilities, meaning they cannot be directly used to jointly sample hidden states. However, the posterior draws generated by `hmm_latent_rng` account for the correlation between hidden states."

charlesm93 (changes requested):
@bbbales2 I think we're almost there. After reading this again, I found a few points which I believe are worth clarifying. Can you make the edits?

bbbales2:
@charlesm93 Edits! Edits! Edits! Weeee

charlesm93 (approved):

PR #192, opened 2021-02-01
[WIP] Feature/math guide

bbbales2:
My first attempt was to write intro docs for all the new stuff in Stan. That started getting long and I was barely making it anywhere so I gave up. Rendered attempt is [here](https://htmlpreview.github.io/?https://github.com/stan-dev/example-models/blob/feature/math-guide/knitr/math-guide/math_guide.html)

Then I tried to write a cheat sheet for all the new stuff in Stan. The idea being that things only needed roughly documented examples and a few words. It's still crazy long, not sure it will fit on a cheat sheet. Rendered cheat sheet is [here](https://htmlpreview.github.io/?https://github.com/stan-dev/example-models/blob/feature/math-guide/knitr/math-guide/math_cheat_sheet.html)

Now I am thinking that my best chance of finishing something is an "anatomy of a Stan function" document. For this, the goal would be to take an existing function and explain the file structure, tests, and add verbal descriptions of all the little different things inside it (and leave the technical description to the doxygen and examples to other functions).

Anyway putting this stuff up to share, not looking for merge.


knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+
+Error checks should be done with the functions in
+[stan/math/prim/err](https://github.com/stan-dev/math/tree/develop/stan/math/prim/err)
+which throw consistently formatted errors.
+
+### Reference types
+
+Any time a function takes a vector or matrix type, it needs to be able
+to handle an expression. Some expressions can be expensive to evaluate, so
+when they are evaluated, they should be saved. This includes any arithmetic
```
t4c1, line 358:
```suggestion
each expression should be evaluated only once. If the result is needed multiple times it should be saved. Expressions include any arithmetic
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+operation, a matrix multiply, or a sum, or something else. Some expressions
+are cheap to evaluate though, any of the Eigen views quality here (transpose,
```
t4c1, line 360:
```suggestion
are cheap to evaluate though, any of the Eigen views qualify here (transpose,
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+The following code will segfault on the construction of `c` because
+by this time the temporary `b` will have been destructed.
+
+```
+Eigen::VectorXd a(5);
+Eigen::VectorXd c = add_zero(a);
+```
+
+`make_holder` can be used to extend the scope of `b` so that it does
+not immediately get destructed.
```
t4c1, line 407:
```suggestion
`make_holder` can be used to extend the lifetime of `b` so that it matches the expression it is used in and does
not immediately get destructed.
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+
+```cpp
+template <typename T>
+auto add_zero(const T& a) {
+  Eigen::VectorXd b = Eigen::VectorXd(a.size());
+  return make_holder([](const auto& l, const auto& r) { return l + r; }, std::move(b));
```
t4c1, line 413: This would not work. The functor accepts two arguments and you are only passing one into `make_holder`.

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+  the tests, just check that an exception is thrown and the exception
+  type is correct.
+
+2. Check that the function returns the expect values for a couple
+suitable inputs.
+
+  For some functions these checks should be more extensive that others,
+  but there should at least be a two or three even in simple cases.
+
+The `prim` tests should only have scalar arithmetic types. No autodiff
```
t4c1, line 458:
```suggestion
The `prim` tests should only have arithmetic types. No autodiff
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+### Arena memory
+
+Reverse mode autodiff in Math requires a huge number of temporaries and
+intermediates to be computed and saved for the reverse pass. There are
+so many allocations that the overhead of `malloc` becomes noticable. To
+avoid this, Math provides its own memory arena. The assumptions of
+the Math arena are:
+
+1. Individual objects allocated in the arena do not need their destructors
+called
```
t4c1, line 491:
```suggestion
called (although there are exceptions)
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+2. All objects allocated on the arena have the same lifetime (that is the
+lifetime of the stack)
```
t4c1, line 493:
```suggestion
lifetime of the stack - until `recover_memory()` is called)
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+
+Because of this, the Math arena is very easy to implement and is very
+high performance (it is a `malloc` that never needs to free).
+
+Functions that use reverse mode autodiff are allowed to save whatever they
+need to the arena. Objects saved here will be available during the reverse
+pass.
+
+There is a utility for saving objects into the arena that do need
+their destructors called. This is discussed undere `make_callback_ptr`.
```
t4c1, line 503:
```suggestion
their destructors called. This is discussed under `make_callback_var`.
```

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+
+`.adj()` and `.adj_op()` return expressions that evaluate to the adjoints
+of the autodiff matrix.
+
+The non-`_op` versions of the functions should be preferred. If the
+code does not compile (usually an error about const-ness) try the `_op`
+versions. This can happen when multiplying the values/adjoints of an AOS
+autodiff type against other matrices.
+
+It is not safe to use `.noalias()` with the AOS value/adjoint expressions.
```
t4c1, line 586: Why not?

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+
+#### `make_callback_var`
+
+`make_callback_var(x, f)` is similar to `reverse_pass_callback`, but it constructs
+an SOA autodiff type from the argument `x`. The function `f` is called on the
+reverse pass similarly to `reverse_pass_callback`.
+
+### Multiple argument scalar types
+
+Pretend for a minute that
```
t4c1, line 606: This looks unfinished.

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+### `forward_as`
+
+### Nested Stacks
+
+### Scoped Stacks
+
+As an aside, in order to get the actual vector result, Eigen defines the
+`.eval()` operator, so `c.eval()` would return an actual `Eigen::VectorXd`.
+Similarly, Math defines the `eval` function which will evaluate an Eigen
+expression and forward anything that is not an Eigen expression.
```
t4c1, line 641: This seems to be in wrong section.

t4c1 (changes requested):
This looks great! I bet it will help any new developers a lot.

knitr/math-guide/math_guide.Rmd
bbbales2: I'm not sure, but I believe it exploded when I did it before. When @SteveBronder did this he added some if statements so the noalias code only ran for SOA. I forget where it was though (I thought multiply but when I looked it wasn't there).

bbbales2:
Judging from your comments @t4c1 you like the longer document? So it's fair to count that as a vote to just finish the longer thing instead of a cheat-sheet style doc?

t4c1:
Well the longer one will be more useful for new devs. Also it looks like 90% finished to me.

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+Math was originally written before C++11. It currently targets C++14. In the near
+future it will transition to C++17. With this in mind, there are many different
+ways to write Math functions. This guide tries to document best practices,
+conventions which not all functions in Math follow, but should be followed for
+new code to keep the code from getting unweildy (the old patterns will be
+updated eventually).
+
+The title contans "Current State" to emphasize that if any information here is
+out of date or any advice does not work, it should be reported as a bug (the
+[example-models](https://github.com/stan-dev/example-models)).
```
SteveBronder, line 28: [todo] move this over to the math site.

SteveBronder:
fyi working on writing some stuff up based on this as well

bbbales2:
@SteveBronder feel free to edit here or move this to where you want to work on it.

Based on what Tadej said, I'm thinking ditch the cheat sheet (but I'll just leave it in for copy-pasting convenience)

knitr/math-guide/math_guide.Rmd
```diff
@@ -0,0 +1,643 @@
...
+```
+
+In this case, there is no generic `T norm(T)` function to write because the
+input argument is actually different than output type. To handle this, Stan
+provides the template metaprogram `return_type_t`. For a given template
+argument `T`, `return_type_t` computes the scalar autodiff return type.
+
+- If `T` is `double`, `return_type_t<T>` is `double`
+- If `T` is `var`, `return_type_t<T>` is `double`
+- If `T` is `Eigen::Matrix<var, Eigen::Dynamic, 1>`, `return_type_t<T>` is `double`
```
gregorp90, line 156: Should return_type_t<T> return var in lines 155 and 156?

bob-carpenter:
Sounds great.  @seantalts and I went through a whole lecture at StanCon Helsinki on how to add a function to the math library.  And we didn't even cover all the doc we had written then.  So this is a big problem.  I also want to say that as long as what's there is correct, it doesn't matter if it's not 100% complete---it'll be an improvement on what we have.

bbbales2:
The new place to work on this is https://github.com/stan-dev/math/pull/2350.

I merged the comments from this thread into that.

PR #199, opened 2021-02-07
Update ARM Ch9 models

andrjohns:
Updates ```ARM/Ch.9``` models with ```_glm``` syntax


rok-cesnovar (approved):
Thank you Andrew! These are all fantastic improvement. Hopefully users start using all the new GLMs.

andrjohns:
Thanks! I hope so as well, there are some pretty clear speed-ups in the models when they're used!

PR #203, opened 2021-02-26
Remove/fix models that dont compile

rok-cesnovar:
This simply removes two models that do not compile and the data for one that had it:
- ARM/Ch.21/finite_populations.stan
- ARM/Ch.21/r_sqr.stan

These have not been compiling for 5+ years (https://github.com/stan-dev/example-models/issues/23, https://github.com/stan-dev/example-models/issues/127) so removing them seems like the best way forward. Otherwise example-models repo is very hard to use for testing Stan-to-C++ compiler regressions.

And fixes three models that required fixes (2 were missing data qualifiers, 1 was using a UDF with a name shadowing Stan Math). 

Other models here compile fine.


bbbales2 (approved):
All good!

Issue #204, opened 2021-02-28
Subtle error in some capture-recapture likelihoods


mbjoseph:
There are some subtle errors in the likelihoods of some capture-recapture models ([Chapter 6 of the Bayesian Population Analysis translations](https://github.com/stan-dev/example-models/tree/master/BPA/Ch.06)). I noticed the errors as a side effect of digging into a separate issues on the Stan Forums: https://discourse.mc-stan.org/t/capture-recapture-model-with-partial-or-complete-pooling/20393/32

Here's an overview of the model that should help clarify what the problem is: 

### State model

These models are two-component finite mixture models, where the components correspond to a partly observed binary latent variable z[i] for individuals i=1, ..., N. The two components are: 

- `z[i] = 0` the individual is not in the population
- `z[i] = 1` the individual is in the population

This latent quantity is modeled as a Bernoulli: `z[i] ~ Bernoulli(omega)`.

### Observation model

There are T sampling occasions, and on each sampling occasion an individual is either detected or not. The observation model is `y[i, t] ~ Bernoulli(p * z[i])` for t=1, ..., T, or summing the number of detections to use a binomial instead, `y[i] ~ Binomial(T, p * z[i])`. 

### Stan implementation

Marginalizing over the discrete state `z` leads to the following integrated observation model: https://github.com/stan-dev/example-models/blob/master/BPA/Ch.06/M0.stan#L29-L38

Essentially, when an individual is observed, we know `z[i] = 1`, and otherwise we have to sum over `z[i] = 0` and `z[i] = 1` in the likelihood.

## The issue

For some of these capture-recapture models, there's an error in the likelihood associated with the case when an individual is in the population but not detected (`z[i] = 1`, and `max(y[i, 1:T]) = 0`). 

The observation model conditional on `z[i]=1` should be `binomial_lpmf(0 | T, p)`, or equivalently `bernoulli_lpmf(y[i] | p)`, where `y[i]` is an integer array of 0's of length T. The following models have a subtle error in that they incorrectly use `bernoulli_lpmf(0 | p)` instead: 

- [Mb: capture effects on detection](https://github.com/stan-dev/example-models/blob/master/BPA/Ch.06/Mb.stan#L53)
- [Mt: time-varying detection](https://github.com/stan-dev/example-models/blob/master/BPA/Ch.06/Mt.stan#L37)
[...]

PR #205, opened 2021-02-28
Fix capture-recapture likelihoods

mbjoseph:
A bug in capture-recapture likelihoods implied only one non-detection for individuals that may or may not be in the population, when in fact we have T non-detections (one for each sampling occasion). Issue described in #204.


ito4303:
Looks good to me.

bbbales2:
@mbjoseph @ito4303 thanks for looking at this. Is this ready to merge?

mbjoseph:
Yes @bbbales2 this is ready to merge - thanks!

PR #206, opened 2021-03-02
Fix broken models and add CI

rok-cesnovar:
- fixes broken models (not compiling): 
   - knitr/bandits/bernoulli-ab.stan
   - knitr/car-iar-poisson/pois_icar.stan
   - knitr/neural-nets/nn.stan

I think the fixes are correct (based on similar models in the subfolders).

Adds Continous integration that will try to compile all modified or added models.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,0 +1,41 @@
+name: Compile models
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+        architecture: 'x64'
+    
+    - uses: actions/checkout@v2
+      with:
+       repository: 'stan-dev/performance-tests-cmdstan'
+       submodules: recursive
+    
+    - uses: actions/checkout@v2
+      with:
+        path: 'example-models-new'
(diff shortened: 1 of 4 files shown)


.github/workflows/compile.yml
```diff
@@ -0,0 +1,41 @@
...
+    
+    - uses: actions/checkout@v2
+      with:
+        path: 'example-models-new'
+        fetch-depth: 0
+    
+    - name: Write modified models
+      run: |
+        cd example-models-new
+        echo "example-models-new/basic_distributions/binormal.stan" > ../test-models.txt
```
rok-cesnovar, line 33: This is just so we always compile at least one model (if changes in PR dont touch models).

bbbales2: Are tests here only running when the test models change?

That doesn't make sense to me. Shouldn't the tests run if the code changes? We do not expect the models to change much I thought.

rok-cesnovar: These tests are for PRs/pushes to example models repo. So we dont add broken models from now on.

rok-cesnovar: They are cloned and compiled/run in other repos (like stanc3 tests will use them here: https://github.com/stan-dev/stanc3/blob/master/Jenkinsfile#L149 once the broken tests are fixed).

bbbales2: Oooo, got it

bbbales2 (approved):
Good!

PR #207, opened 2021-04-05
rename quantile udf function to avoid name clash with math

adamhaber:
In https://github.com/stan-dev/stanc3/pull/870 we expose a quantile function, which caused a named clash with 4 example models.

This fixes the name clashes by renaming the `quantile` function in the models to `quantile_user_defined`.


rok-cesnovar (approved):
Looks good.

PR #210, opened 2021-09-17
Run canonicalizer on all models

WardBrian:
This runs `stanc --print-canonical` on all files. This removes a lot of deprecated syntax, in addition to auto-formatting all the code.

This is a draft because I am still looking it over myself and some changes to stanc3 are still being cleared to make the formatting prettier (https://github.com/stan-dev/stanc3/pull/970)

Fixes:
- `array[]` syntax
- Replaces `<-`
- Replaces `if_else`
- Replaces increment_log_prob
- Fixes some names of functions, particularly the suffixes

Does not fix:
- ODE solver changes


BPA/Ch.06/M0.stan
```diff
@@ -1,45 +1,46 @@
...
+      target += bernoulli_lpmf(1 | omega) + binomial_lpmf(s[i] | T, p);
+    // s[i] == 0
+    
+    else 
+    
+      target += log_sum_exp(bernoulli_lpmf(1 | omega)
+                            + binomial_lpmf(0 | T, p),
+                            bernoulli_lpmf(0 | omega)); // z[i] == 0
+  
+  // ^^^: z[i] == 1
```
WardBrian, line 39: @nhuurre - related to https://github.com/stan-dev/stanc3/pull/894, do you think there is any way we could handle this better when auto-formatting?

nhuurre: The bad formatting here is 
1. too much space around `else` 
2. `s[i] == 0` comment on the wrong side of `else`
3.  `z[i] == 1` comment pushed out completely

(1) is probably not too bad, just requires some tweaking `pp_recursive_ifthenelse`. (2) is annoying because the AST does not record keyword locations. Could either change the AST or record keywords in the comment list (like commas). (3) is a case of comments inside binary operator expressions being skipped. Probably easy to fix, the printer code should be very similar to the code for expression lists in `pp_list_of`. (but again, needs a hack to find the exact location of the operator token)

Btw, manually adding braces and moving the `z[i]` comment to the next line makes the auto-format produce
```stan
  for (i in 1 : M) 
    if (s[i] > 0) {
      // z[i] == 1
      target += bernoulli_lpmf(1| omega) + binomial_lpmf(s[i]| T, p);
    } else {
      // s[i] == 0
      target += log_sum_exp(bernoulli_lpmf(1| omega) + binomial_lpmf(0| T, p)
                            // z[i] == 1
                            , bernoulli_lpmf(0| omega)); // z[i] == 0
    }
```

WardBrian: I think it is worthwhile to make edits to the AST to retain more location information as needed. It's probably not bad to have around anyway and aligns with the "AST is intended to have a direct 1-1 mapping with the syntax, so there are things like parentheses being kept around" goal of stanc3, IMO. I also think it's better to have in the AST than in the comment list like we're currently doing for commas, but I can be persuaded either way.

I agree that your formatting with the braces is preferrable, but I think this is also a decent example of the kind of thing someone has written 'in the wild', and it would be good to support as much as possible. Would you be able to take a look at this some more? Even just fixing the "easy" parts you outlined would be a great plus

WardBrian:
@rok-cesnovar I'm going to manually go through this PR now, searching primarily for `// ^^^ ` comments and manually moving them to better positions. After that, want to scroll through as well to make sure it all looks reasonable here?

rok-cesnovar:
Yeah, tag me once you think its good to go.

WardBrian:
@rok-cesnovar I'm starting my manual comb through now (ran again with https://github.com/stan-dev/stanc3/pull/1055), unless I find anything obviously wrong I think this will be good to go.

WardBrian:
Finally had the chance to scroll through the entire set of changes. All look good to me. I'd 'approve' if I could, but they don't let you do that as the author.

@rok-cesnovar if you get a chance to take a look please do!

rok-cesnovar:
Wrote down as todo for tomorrow.

rok-cesnovar (approved):
Great stuff!

PR #212, opened 2021-11-19
Auto-format and canonicalize the example-models

rok-cesnovar:
Just ran `stanc --auto-format --print-canonical` to make the formatting consistent and make all the models us the new array syntax.


WardBrian:
I’ll scroll through next week. Is this substantially different to #210? (not sure if you can diff them directly)

rok-cesnovar:
Oh man, I forgot about #210. Will close this then.

PR #214, opened 2022-01-24
update case study of hierarchical models for binary trials 

mitzimorris:
Updated models and notebook for Stan case study "Hierarchical Partial Pooling for Repeated Binary Trials"

- use affine transform for non-centered hierarchical logit model
- use include directives for data, transformed data, and generated quantities code shared across all models
- rework Rmd notebook so that it runs with CmdStanR

### DIFF
--- a/knitr/pool-binary-trials/pool-no-pool.Rmd
+++ b/knitr/pool-binary-trials/pool-no-pool.Rmd
@@ -1,11 +1,10 @@
----
-title: "Hierarchical Partial Pooling for Repeated Binary Trials"
-author: "Bob Carpenter"
-date: "29 January 2016"
-output: 
-  html_document: 
-    theme: readable
----
+## Hierarchical Partial Pooling for Repeated Binary Trials in CmdStanR
+
+*Bob Carpenter* <br>
+29 January 2016
+
+updated for CmdStanR by Mitzi Morris, 19 January 2022
+
 
 ### Abstract
 
@@ -18,13 +17,45 @@ unrelated, and (c) partial pooling, where the similarity among the
 items is estimated.  We consider two hierarchical models to estimate
 the partial pooling, one with a beta prior on chance of success and
 another with a normal prior on the log odds of success.  The note
-explains with working examples how to (i) fit models in RStan and plot
-the results in R using ggplot2, (ii) estimate event probabilities,
-(iii) evaluate posterior predictive densities to evaluate model
-predictions on held-out data, (iv) rank items by chance of success,
-(v) perform multiple comparisons in several settings, (vi) replicate
-new data for posterior p-values, and (vii) perform graphical posterior
-predictive checks.
+explains with working examples how to
+
+- (i) fit models in [CmdStanR](https://mc-stan.org/cmdstanr) and plot the results in R using ggplot2,
+- (ii) estimate event probabilities,
+- (iii) evaluate posterior predictive densities to evaluate model predictions on held-out data, (iv) rank items by chance of success,
+- (v) perform multiple comparisons in several settings,
+- (vi) replicate new data for posterior p-values, and (vii) perform graphical posterior predictive checks.
+
+
+### Packages used in this notebook
+
+In this notebook, we will use the [CmdStanR interface](https://mc-stan.org/cmdstanr/).
+
+```{r comment=NA}
+#library(devtools);
+#if(!require(cmdstanr)){
+#    devtools::install_github("stan-dev/cmdstanr", dependencies=c("Depends", "Imports"))
+#}
+options(warn=-1)
+options(repr.plot.width=8, repr.plot.height=8)
+library(tidyverse)
+library(posterior)
+library(cmdstanr)
+set_cmdstan_path("/Users/mitzi/.cmdstan/cmdstan-2.28.2")
+```
+```{r echo=FALSE}
+theme_jupyter <- function() {
+  theme_grey() + 
+  theme(text = element_text(size=18),
+        plot.title = element_text(size=24),
+        plot.subtitle = element_text(size=20),
+        axis.title.x = element_text(size=20),
+        axis.title.y = element_text(size=20),
+        strip.text.x = element_text(size=18),
+        strip.text.y = element_text(size=18)
+       );
+}
+```
+
 
 ## Repeated Binary Trials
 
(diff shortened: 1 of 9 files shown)


mitzimorris:
This PR updates the Rmd notebook and the  HTML
A subsequent PR will break this case study into multiple parts and create Jupyter notebooks.

WardBrian:
To actually update the one people see I think there needs to be a separate PR to https://github.com/stan-dev/stan-dev.github.io, right?

I’ll take a look at this tomorrow

mitzimorris:
If you OK this PR, that should be sufficient review for the stan-dev.github.io pages as well.

mitzimorris:
can we turn off compilation for the include fragments?  if so, how?

WardBrian:
I think @rok-cesnovar put it together. My guess is provably naming them something other than .stan would work, but then you lose syntax highlighting etc

mitzimorris:
@rok-cesnovar  - the include files are in a directory named "include"  - can we do something with this information?
for this case study, using includes makes a lot of sense. also, it's good to have more examples of how this works somewhere.

knitr/pool-binary-trials/pool-no-pool.Rmd
```diff
@@ -160,33 +216,27 @@ due to numerical arithmetic issues.
 
 Assuming each player's at-bats are independent Bernoulli trials, the
 sampling distribution for each player's number of hits $y_n$ is modeled as
-
-\[
+$$
```
WardBrian, line 219: This is a minor detail, but generally using `\[ \]` is preferred over `$$ $$`. The later is a pure-TeX shorthand which can cause some issues with LaTeX macros. Not sure it matters for us, but just thought I'd note. [More on Wikibooks](https://en.wikibooks.org/wiki/LaTeX/Mathematics#Mathematics_environments)

knitr/pool-binary-trials/pool-no-pool.Rmd
```diff
@@ -238,62 +287,77 @@ vectorized form can be up to an order of magnitude or more faster in
...
-The actual Stan program in <code>pool.stan</code> has many more
+The actual Stan program in <code>pool.stan</code> includes many
 derived quantities that will be used in the rest of this note; see
 the appendix for the full code of all of the models discussed.
 
-We start by loading the RStan package.
+We must first instantiate a CmdStanModel object then call its <code>sample</code> method.
+To do this we first call the `cmdstan_model()` function
+which instantiates a CmdStanModel object from a file containing a Stan program.
+Because these models contain a number of [include directives](https://mc-stan.org/docs/reference-manual/includes-section.html),
```
WardBrian, line 297:
```suggestion
Because these models contain a number of [include directives](https://mc-stan.org/docs/reference-manual/includes.html), 
```

With `-section` it will link to the 2_27 version (at least until we fix that redirect)

knitr/pool-binary-trials/pool-no-pool.Rmd
```diff
@@ -869,50 +964,62 @@ random-walk Metropolis samplers (Papaspiliopoulos et al. 2003).  This
...
+This decouples the sampling distribution
+for $\alpha^{\mathrm{std}}$ from $\mu$ and $\sigma$, greatly reducing
+their correlation in the posterior.  
 
-\[
-p(y_n \, | \, \alpha^{\mathrm{std}}_n, \mu, \sigma, K)
-\ = \
-\mathsf{BinomialLogit}(K_n, \ \mu + \sigma \, \alpha_n).
-\]
+Prior to Stan 2.19, a Stan implementation directly encoded the math.
```
WardBrian, line 988:
```suggestion
Prior to Stan 2.19, a Stan implementation directly encoded the math needed to reparameterize.
```

WardBrian (commented):
A few comments on the Rmd file. The reorganization of the stan files using `#include` looks good

rok-cesnovar:
Feel free to ignore the testing errors, will fix the GHA script later this week to ignore the files you mentioned.

knitr/pool-binary-trials/pool-no-pool.Rmd
mitzimorris: yes, but only `$$` works in Jupyter notebooks.

knitr/pool-binary-trials/pool-no-pool.Rmd
mitzimorris: added more prose - hoping to make the use of this clearer.

knitr/pool-binary-trials/pool-no-pool.Rmd
WardBrian: This still needs to be changed (I think you changed a different copy of the link which I hadn't caught)

WardBrian:
It doesn't let me comment directly on the rendered HTML, but here are some thoughts on the way the output looks:

- Can we hide the 'A newer version of CmdStan is available. See ?install_cmdstan() to install it.' message? If possible it would also be great to hide some (if not all) of these setup messages such as the cmdstan path
- There's some strange formatting around the code blocks at the end. Some of the titles seem to stretch onto the next line while still in `typewritertext` 
- I'm not sure how the html is built from the Rmd in this case - would it be possible to get syntax highlighting of the Stan code?

mitzimorris:
> Can we hide the 'A newer version of CmdStan is available. See ?install_cmdstan() to install it.' message? If possible it would also be great to hide some (if not all) of these setup messages such as the cmdstan path

I think this is an CmdStanR bug.  I actually have CmdStan 2.28.2 installed.  I also have a version of develop in my `.cmdstan` directory and I named that dir `cmdstan-develop` -  CmdStanPy ignores it, but CmdStanR wanted to use that and not 2.28.2.  @rok-cesnovar - what are the rules for CmdStanR?

knitr/pool-binary-trials/pool-no-pool.Rmd
mitzimorris: yes, it would be good to be able to run these in jupyter, either in Python or R.

I used jupyter to update this notebook, so I hit this right off the bat.
awhile back I installed juptext on my machine and I prefer jupyter to RStudio - https://mc-stan.org/users/documentation/case-studies/jupyter_colab_notebooks_2020.html#challenge-putting-the-r-in-jupyter

With juptext and IRkernal, I can run the notebook while keeping it in .Rmd format which means that it's easy to edit via the browser of in emacs.  but that's just me.

WardBrian:
It mentions an environment variable to disable it for now

mitzimorris:
I just converted it to a draft because there's more work to be done here.

will investigate formatting issues.

it would definitely be nice to get Stan syntax highlighting into the HTML
and I'm pretty sure we can hack the .Rmd to make that happen.

WardBrian:
What program is used to ultimately create the HTML? Pandoc?

rok-cesnovar:
> @rok-cesnovar - what are the rules for CmdStanR? 

We sort the folder in .cmdstan, not sure why -develop would be ahead though. Will check.

PR #226, opened 2024-02-08
Add JSON data files

WardBrian:
This was done using `cmdstanr::write_stan_json(rstan::read_rdump(file_R), file_json)`. The primary goal is to allow us to use these files in the cmdstan performance tests to avoid warnings about RDump, but they also serve as paired examples of the two data formats.


mitzimorris:
the method seems sound.
did you run the performance tests using the JSON data?

WardBrian:
I just ran them locally. They don't use all the models here (really only the ones in `bugs_examples/` and `regression_tests/`), but seem just as happy with the new files as the old

mitzimorris (approved):
the method is sound and has been tested locally - good enough for me!

PR #228, opened 2024-08-03
Python scripts to BYM12 case study

farhanreynaldo:
The following PR is a draft on `Fitting the Model to the Scotland Lip Cancer Dataset` and `Bigger data: from 56 counties in Scotland to 1921 census tracts in New York City` sections. 

There are several issues need to be addressed. 
* The BYM2 model cannot be directly translated from R to Python because as far as I know, there is no equivalent INLA library in Python. 
* The raw number of 2001 events plot has different value scale compared to the R implementation. The Python plot has a range of value from 0 to 300, while the R plot has 0 to 15.
* There is an exception for BYM model in Python, however the summary still produced.


bob-carpenter:
Pinging @mitzimorris, who wrote the original ICAR case study.

If you want a dependency on a low-profile package, there's https://github.com/al-jshen/pyinla, but it looks like it just wraps the R version, so probably not a good idea.

jupyter/car-iar-poisson/car-iar-poisson.ipynb
```diff
@@ -0,0 +1,1310 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "                                                                                                                                                                                                                                                                                                                                "
```
mitzimorris, line 131: before checking in a jupyter notebook, please reload and delete all outputs.  this makes it much easier to maintain and for others to use.

jupyter/car-iar-poisson/car-iar-poisson.ipynb
```diff
@@ -0,0 +1,1310 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import cmdstanpy\n",
+    "from cmdstanpy import CmdStanModel\n",
+    "import numpy as np\n",
```
mitzimorris, line 18: it would be good to list the Python packages this notebook will be using.

jupyter/car-iar-poisson/car-iar-poisson.ipynb
farhanreynaldo: Noted, I already removed the outputs.

jupyter/car-iar-poisson/car-iar-poisson.ipynb
farhanreynaldo: I have added the Python packages in the markdown block.

PR #229, opened 2025-03-03
sum_to_zero_vector case study

mitzimorris:
Transferring the contents of github repo:  https://github.com/mitzimorris/sum_to_zero_vector  to this repo.

This case study introduces the `sum_to_zero_vector`.  It demonstrates a simple workflow for evaluating performance of different ways to impose a sum-to-zero constraint on a parameter vector.

The HTML file is self-contained.  To re-render the HTML, this requires the `stan-dev/quarto-config` repo for the Stan website styling.


mitzimorris:
hi @spinkney and @WardBrian  - not sure if we need reviews to add case studies, but I would appreciate any feedback you might have, if you have time.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1558 @@
...
+
+* In the `transformed parameters` block, we apply the constraining transform.
+
+```stan
+  vector[N_connected] phi = zero_sum_components_lp(phi_raw, component_idxs, component_sizes);
+```
+
+The constraining transform is broken into two functions:
+
+* function `zero_sum_constrain_lp`, the actual constraining transform, which corresponds directly
```
WardBrian, line 1334: A couple notes on this:

1. This should probably be a `_jacobian` function (another new 2.36 feature), or, because there is no jacobian adjustment, just a normal function (probably worth mentioning the linearity of the transform here)
2. This function will, [most likely](https://github.com/stan-dev/stanc3/pull/1494), be available built-in in 2.37 under the names `sum_to_zero_constrain` and equivalently `sum_to_zero_jacobian`. Not sure if it's worth mentioning that here or doing a edit later

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1558 @@
...
+using the PRNG equivalent of the model likelihood.
+
+* Because the modeled data `pos_tests` is generated according to the Stan model's likelihood,
+the model is a priori well-specified with respect to the data.
+
+* Because the true parameters are defined in the `generated quantities` block,
+each sample provides a datasets from a different set of regression covariates
+and with different amounts of per-demographic data.
+
+The full data-generating program is in file `stan/gen_binomial_4_preds.stan`.
```
WardBrian, line 317: (assuming this will end up on e.g. the stan website) should these file references be a link?

WardBrian (approved):
I only have one real comment, otherwise this looks great!

Will you also open a PR to add the rendered version to the website?

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+
+sim_data_lg = gen_mod.sample(data=gen_data_lg,
+                          iter_warmup=1, iter_sampling=1, chains=1, seed=45678)
+
+sim_data_tiny = gen_mod.sample(data=gen_data_tiny,
+                          iter_warmup=1, iter_sampling=1, chains=1, seed=45678)
+```
+
+##### Examine the set of generated data-generating params and resulting dataset.
+
```
spinkney, line 419: Add sentence

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+```{python}
+print(f'Small dataset: N = {gen_data["N"]}, strata = {strata}, expected obs per demographic {gen_data["N"] / strata}')
+print(f'Large dataset: N = {gen_data_lg["N"]}, strata = {strata}, expected obs per demographic {gen_data_lg["N"] / strata}')
+for var, value in sim_data.stan_variables().items():
+    print(var, value[0]) if isinstance(value[0], np.float64) else print(var, value[0][:10])
+```
+
+What is the distribution of the observed number of tests per demographic?
+
```
spinkney, line 428: I'd prefer an actual table here with the information and the code secondary just to see how you did it.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+```stan
+  vector[N_eth] pct_eth = dirichlet_rng(rep_vector(1, N_eth));
+  for (n in 1:N_eth) {
+    beta_eth[n] = std_normal_rng();
+  }
+```
+
+However, this can result in very unbalanced datasets, in which case it is best to
+generate another dataset and continue.
+
```
spinkney, line 460: what does the following code do vs the block after? I'm being a bit picky but it will be much clearer to the reader.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+    'pct_edu': sim_data.pct_edu[0],
+    'beta_edu':sim_data.beta_edu[0]
+}
+true_params
+```
+
+#### Model 1: `sum_to_zero_vector`
+
+This model is in file [binomial_4_preds_ozs.stan](https://github.com/stan-dev/example-models/tree/master/jupyter/sum-to-zero/stan/binomial_4_preds_ozs.stan).
+
```
spinkney, line 554: I suggest showing at least the parameters block for the `sum_to_zero_vector`. For the below python code, I suggest putting all the following code in one block.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+```{python}
+a_seed = binomial_ozs_fit.metadata.cmdstan_config['seed']
+```
+
+```{python}
+binomial_ozs_fit_lg = binomial_ozs_mod.sample(data=data_large, parallel_chains=4, seed=a_seed)
+```
+
+#### Model 2:  Hard sum-to-zero constraint
+
```
spinkney, line 571: Highlight what has changed from the previous Stan code. I suggest putting all the python into one block. This goes for the following section as well.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+```
+```{python}
+binomial_soft_fit = binomial_soft_mod.sample(data=data_small, parallel_chains=4, seed=a_seed)
+```
+```{python}
+binomial_soft_fit_lg = binomial_soft_mod.sample(data=data_large, parallel_chains=4, seed=a_seed)
+```
+
+#### Runtime performance
+
```
spinkney, line 600: Why even have this section if there isn't a table to show?

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
```diff
@@ -0,0 +1,1582 @@
...
+soft_total_time_lg = 0
+for i in range(len(soft_fit_time_lg)):
+    soft_total_time_lg += soft_fit_time_lg[i]['total']
+soft_fit_lg_summary['ESS/sec'] = soft_fit_lg_summary['ESS_bulk']/soft_total_time_lg
+
+large_data_fits_summary = pd.concat([ozs_fit_lg_summary, hard_fit_lg_summary, soft_fit_lg_summary])
+```
+
+**Eth**
+
```
spinkney, line 675: What is this? What does all the code do? Don't assume your readers know python that well!

spinkney (changes requested):
I mostly added comments that hopefully helps the reader to see the differences and the results more clearly. I'm happy that you put this together and I'm happy to chat about any of the comments if you'd like.

mitzimorris:
rewrote the case study per @spinkney comments - which were absolutely spot on!

- the essential comparisons are presented as tables
- data processing is summarized, but not shown

should I omit the ipynb or update to match?

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
mitzimorris: rewrote - now stands as:

Because the data-generating parameters and percentage of observations per category are generated at random,
some datasets may have many low count or no count strata, or random effects near zero across several categories
and will therefore be pathologically hard to fit.
Using a specified seed for the data-generating program avoids this problem.

bob-carpenter:
> Because the data-generating parameters and percentage of observations per category are generated at random,
some datasets may have many low count or no count strata, or random effects near zero across several categories
and will therefore be pathologically hard to fit. Using a specified seed for the data-generating program avoids this problem.

This is basically saying that your model fails SBC as specified.  One way to fix it would be to (a) tighten the priors so they're not so extreme, and (b) avoid simulating empty strata by generating the fill level uniformly in (0.5, 1) rather than uniformly in (0, 1).  That would probably be preferable than a disclaimer that says SBC fails.

I wouldn't say "data-generating parameters"---the parameters are arguments to the distributions that generate data, but they don't generate data themselves.   I would also prefer "zero count" to "no count".  I would also prefer "varying effects" to "random effects" for all the reasons Andrew mentions.

How about:

> Because the parameters are simulated randomly from their priors and the number of observations per category are also simulated randomly, some data sets may have many low or zero count strata.   This will lead to varying effect parameters with near-zero values, which are on the boundary of parameter space, and hence very challenging to fit.  Cherry-picking a seed that works avoids these issues.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
bob-carpenter: Thanks---that looks good.

jupyter/sum-to-zero/sum_to_zero_evaluation.qmd
bob-carpenter: @spinkney: What are you asking for here?  Documentation within the code?  A description summarizing what it does on the outside?  

I would assume the readers know Python pretty well.  Otherwise, the intro to Python will overwhelm the intro to sum-to-zero.  If you do want to teach people bits of Python, you should have a model reader in mind or it will be very hard to decide where to stop with expounding on the Python code.  Is that someone who knows Python syntax but not any libraries like numpy or matplotlib?  Or someone even less skilled who doesn't understand python scoping or control flow?  Maybe they don't know what `+=` means or that `range(N)` is from 0 to `N-1`?

mitzimorris:
@spinkney, @bob-carpenter - ready for review

PR #230, opened 2025-03-25
Add comments

ito4303:
Added explanations on the generation of posterior predictive distribution


bob-carpenter:
I'm going to merge this, but I'm a bit curious, because the comments say:

```stan
// This code generates the posterior predictive distribution
// unconditioned on the observation
```

A posterior predictive distribution is by definition conditioned on some observations, it's $p(\tilde{y} \mid y)$ where $\tilde{y}$ is new data and $y$ is observed data.

bob-carpenter:
P.S.  Thanks for continuing to maintain this.  We still use a lot of these models for bulk testing and people still use them as the basis for models they develop.

PR #231, opened 2025-03-26
Correct comments

ito4303:
delete "unconditioned on the observation" in the comments


bob-carpenter (approved):
Looks good---thanks for updating.

